PR #69, opened 2015-01-27
Add support for PHP code cleanup using PHPTools ver. 1.15.6124 Preview

ikappas:
Hi Steve,

I can confirm that the PHPTools guys have fixed the error traversing the projectItem.Collection and now CodeMaid cleanup works like a charm with their latest PHPTools (ver. 1.15.6124) that is available as a Preview at:
http://www.devsense.com/products/php-tools/download

To test this you need to install PHPTools (available as a 30 day trial) in the Visual Studio Experimental instance. Instructions can be found here:
http://support.devsense.com/viewtopic.php?f=21&t=1407

Digging: not supported because they do not expose projectItem.FileCodeModel.
Formatting: I will have a look at it soon


codecadwallader:
Looks very good - thank you for your contribution!  Any objection to me referencing you as the contributor on blog posts, etc?

ikappas:
Great! No problem :)

PR #91, opened 2015-04-04
fixes #90: requires whitespace after list prefix; only allow up to 3 cha...

### LINKED ISSUE #90: Hyperlink in xml doc is formatted incorrectly
To reproduce, simply type in following xml doc:

    /// <summary>
    /// http://foo
    /// </summary>

After formatting comment it will become:

    /// <summary>
    /// http: //foo
    /// </summary>

wangzq:
...racters as list prefix if it is one of "-=*+"

### DIFF
--- a/CodeMaid.UnitTests/Formatting/XmlFormattingTests.cs
+++ b/CodeMaid.UnitTests/Formatting/XmlFormattingTests.cs
@@ -256,5 +256,14 @@ namespace SteveCadwallader.CodeMaid.UnitTests.Formatting
                 XmlSplitSummaryTag = false
             });
         }
+
+        [TestMethod]
+        [TestCategory("Formatting UnitTests")]
+        public void XmlFormattingTests_Hyperlink()
+        {
+            var input = "<summary>" + Environment.NewLine + "http://foo" + Environment.NewLine + "</summary>";
+            var expected = input;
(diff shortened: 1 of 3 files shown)


willemduncan:
I can see a good reason for the "requires whitespace after list prefix" change (which fixes the hyperlink problem), but there is more changed here. Can you please explain the following changes?

1. Why limit the prefix characters to maximum of 3 when if it is one of "-=*+"?
2. Why have a `TrimEnd()` on the list prefix? Maybe I want extra spacing for outlining, eg:
````
1)  Something
10) Something else
````

wangzq:
I am not on the computer with the code so following info is based on my memory now:

1. if I didn't limit the number of prefix characters then an existing unit test will fail, as it is using "--------------" as horizontal separator line; the number 3 is arbitrary here, please feel free to change it.

2. Currently without TrimEnd if you format it a few times then each time a new space is inserted; that said, I think you are right that I should find out the place where the new space is being inserted and remove this TrimEnd instead; I will do this later if you prefer.

willemduncan:
I copied the unit tests but altered the regular expression logic in a different way: The list prefix now includes its whitespace, and no more whitespace is added in the formatting logic. This means there is no need for either the `{1,3}` lenght requirement or the `TrimEnd()` call. Thanks for the help in finding and fixing this issue!

Issue #106, opened 2015-06-15
CodeMaid treats commented out code as regular comments and makes it unusable after uncommenting


shytikov:
Hi,

Minor, but really annoying issue. Sometimes during development it's needed to hide some pieces of code from execution. The straightforward way to do it — just comment them out. By whole blocks of methods or even classes.

But.

If developer runs `CodeMaid`'s cleanup to make rest of the code pretty the addon will re-arrange comments with code hidden in the. For example `CodeMaid` will run word-wrap on longer lines, and this will make code uncompilable when uncommented.

I think `CodeMaid` should check is this text comment or piece of code. And in case of last — preserve formatting.

codecadwallader:
Detection of code vs. comment is challenging, so we follow some basic rules.  If there are two slashes, immediately followed by a character - we assume that is code when it comes to comment formatting.  If there are ever four slashes (StyleCop's recommended practice http://stylecop.soyuz5.com/SA1626.html) - we also assume that is code.

Wrapping should only be occurring if it is unable to detect the comment could potentially be code.

Can you provide a code sample that is formatting when you don't think it should?

monze:
I have an example. Just created it, but we use this approach many places in our source code. We have some debug helping implementation in the code, that can be enabled to trace an error. As default it is comment out. The code is just created as test, and will not work, and serves only as example.

```
        /// <summary>
        /// Get the values of the xPath
        /// </summary>
        /// <param name="xmlDocument">The xmlDocument</param>
        /// <param name="xPath">The xpath path</param>
        /// <returns>The value, or string.Empty</returns>
        public static string GetValue(XmlDocument xmlDocument, string xPath)
        {
            string value = string.Empty;

            try
            {
                // Dummy code - does not work
                value = xmlDocument.ParseXpath(xPath).GetValue;
            }
            catch (Exception ex)
            {
                Debug.Fail(ex.Message, ex.ToString());
            }

            // The document is retrieved from the Internet. Save the document to d:, to debug why
            // the xPath failed, just uncomment the lines below.
            // string fileName = "D:/GetValueDocument.xml";
            // xmlDocument.Save(fileName);

            // return the value
            return value;
        }
```

The code as it is, is StyleCop valid, and we have some implementation, we can uncomment to help debug the situation.
```
            string fileName = "D:/GetValueDocument.xml";
            xmlDocument.Save(fileName);
```

This is just a short implementation, but could easy be several lines long, dependent of the situation.
[...]

PR #109, opened 2015-06-25
Updated comment formatting logic to address some issues.

willemduncan:
This pull request includes fixes for #90 and #106. Added extra unit tests to test these cases, and found no regression on any of the previous functionality.

Sadly I had a bit of a fight with Visual Studio version differences and whitespace handling, this marks some files as changed even though they really aren't.


codecadwallader:
Thanks Willem.  It looks good from my quick code scan.  I'll pull it down this weekend and look to merge.

Issue #63, opened 2015-01-11
Problem with formatting lists


bartlomiejwolk:
Hi, this is how CodeMaid formats lists in my code:

![xml_list](https://cloud.githubusercontent.com/assets/325348/5694552/9badd97e-9966-11e4-9e8e-668d1a031e56.png)
Tags inside list are just merged together instead of creating a structure.

And this is my CodeMaid config:
![codemaid_config](https://cloud.githubusercontent.com/assets/325348/5694555/b84aa8d2-9966-11e4-98f2-6965c94a9ddb.png)

Is this an expected behavior?

codecadwallader:
We look for one of several different types of punctuation in order to recognize that the content is intended to be an item within a list.  Currently we don't use an XML tag as the basis for a list item.

Some list indicators that are recognized:
    -
    =
    *
    +
    a:
    1.

Does that help?

bartlomiejwolk:
Yes it does although it would be nice if xml list tags were supported.

willemduncan:
Is there a reason for using this syntax? I don't see Visual Studio recognizing the format (they're not standard XML doc tags).

codecadwallader:
Agreed, added to backlog here: https://trello.com/c/8PjGOadD

PR #166, opened 2015-10-18
Fix splitting of XML comment tags which contain a single <para> element.

reima:
This PR makes CodeMaid format the XML comment…

```
<test><para>Lorem ipsum.</para></test>
```

…like this:

```
<test>
<para>Lorem ipsum.</para>
</test>
```

Currently, CodeMaid would format it as:

```
<test>
<para>Lorem ipsum.</para></test>
```

I also added a unit test for this case.

### DIFF
--- a/CodeMaid.UnitTests/Formatting/XmlFormattingTests.cs
+++ b/CodeMaid.UnitTests/Formatting/XmlFormattingTests.cs
@@ -205,6 +205,19 @@ namespace SteveCadwallader.CodeMaid.UnitTests.Formatting
 
         [TestMethod]
         [TestCategory("Formatting UnitTests")]
+        public void XmlFormattingTests_SplitsTagsWhenItContainsElementOnSeparateLine()
+        {
+            var input = "<test><para>Lorem ipsum.</para></test>";
+            var expected =
+                "<test>" + Environment.NewLine +
+                "<para>Lorem ipsum.</para>" + Environment.NewLine +
+                "</test>";
+
+            CommentFormatHelper.AssertEqualAfterFormat(input, expected);
+        }
+
+        [TestMethod]
+        [TestCategory("Formatting UnitTests")]
         public void XmlFormattingTests_TagNameKeepCase()
         {
             var input = "<Summary></Summary>";
--- a/CodeMaid/Model/Comments/CommentFormatter.cs
+++ b/CodeMaid/Model/Comments/CommentFormatter.cs
@@ -156,7 +156,10 @@ namespace SteveCadwallader.CodeMaid.Model.Comments
             var xml = line as CommentLineXml;
             if (xml != null)
             {
-                return ParseXml(xml, indentLevel);
+                ParseXml(xml, indentLevel);
+
+                // XML lines always introduce a line wrap.
+                return true;
             }
             else
             {
@@ -315,8 +318,7 @@ namespace SteveCadwallader.CodeMaid.Model.Comments
             return false;
         }
 
-        /// <returns><c>true</c> if line fitted on single line, <c>false</c> if it wrapped on multiple lines.</returns>
-        private bool ParseXml(CommentLineXml line, int indentLevel = 0)
+        private void ParseXml(CommentLineXml line, int indentLevel = 0)
         {
             // All XML lines start on a new line.
             if (!_isFirstWord)
@@ -388,8 +390,6 @@ namespace SteveCadwallader.CodeMaid.Model.Comments
             }
 
             Append(line.Closetag);
-
-            return tagOnOwnLine;
         }
 
         /// <summary>


codecadwallader:
@reima Thank you for the contribution, and especially for including a unit test going forwards.  I'm not sure if the change to ParseXml to no longer return a result has any other side effects.  I didn't find any in my initial testing, but will pass to @willemduncan who owns the comment formatting logic.

willemduncan:
If I recall correctly the `ParseXML` returned a boolean, which is used to decide whether the XML tag fits on a single line or not. This is used in the automatic line splitting on root level XML tags if the content flows to more than a single line. At least, that's what I think it does.

Anyway, I see in the changed code that "`// XML lines always introduce a line wrap.`", which is not always true. First off self closing tags do not force a linewrap, eg `<see cref="..." />`. Second, a short `<c>code</c>` tag or `<p>para</p>` tag should also not wrap. I do not know if we already unit test those situations, but they are important to the wrapping logic.

Sadly I'm currently not in a position to test the impact of this PR, I can pick this up next week the earliest.

reima:
@willemduncan My reasoning behind this change was that the first line of ParseXML says "`// All XML lines start on a new line.`". Starting on a new line means a line wrap (right at the start).

As far as I understand the comment parsing logic, `CommentLineXml` objects are only generated for XML comments which should be on a separate line (e.g. tags like `<code>`, `<para>`, or top-level tags). Inline tags like `<see>` or `<c>` don't create a new `CommentLineXml` object and as such are not affected by my changes.

willemduncan:
Good reasoning, that might very well be true!

willemduncan:
Ok so the whole thing kept me thinking anyway, so I double checked what's up (mostly for myself, it's been a while since I touched or thought about this code). What happens in the parsing: 

- An XML node is parsed into an `XMLCommentLine`, which then recursively parses it's children. 
- If the XML node is a literal node (for now only `<code>` elements), the content is copied and not formatted.
- All other cases the child nodes are parsed and added to a `CommentLine` under the XML line.
- If a special node is found, eg `<p>`, `<para>` and `<code>`, they are parsed into a new `XMLCommentLine`, and the whole thing starts again.

This means that part of the decision for whether an XML comment line should break or not is already done on the parsing level, and not on the writing level. Not a very good thing, design wise. From this comes the bit where
```
// All XML lines start on a new line.
```
Which is for the current flow correct, but ultimately wrong. We should work to a system where all XML tags are parsed into XML comment lines, and only at the "writer" level decide whether they can stay inline or not. This will allow proper handling of some other edge cases as well, e.g. prevent splitting of the long `<see cref="..." />` tags, and thus makes the `bool` return relevant.

So the problem I have now: Your fix is correct, does what it should and has no side effects as far as I see. However I think removing the return type from the output function is moving away from the ultimate desired situation. Crazy as it would be, I feel a simple `return true` would for now be more fitting. I'm of course open for suggestions!

reima:
Excellent analysis and summary of the problem. I agree on all points. If there are no objections, I will make your suggested change so we can move forward with this PR. The parsing/formatting design flaw can be fixed later.

To give a little more context: I noticed the formatting bug while working on #63 ([card 220](https://trello.com/c/8PjGOadD/220-support-xml-list-tags)). I tried to solve this by adding `<list>`, `<item>`, `<description>` etc. to the list of special nodes which should be on a separate line. Now it seems this would be a further move in the wrong direction. Should I postpone working on #63 until the parsing/formatting is cleaned up?

willemduncan:
I have started and abandoned cleaning and rebuilding the whole process of parsing and formatting a few times. Every time I got stuck on proper handling of XML, it's quite tricky to get right. Right now it's only a rough idea in my mind, no actual code beyond some proof of concept fragments exist. As far as I see a future version of the formatter should be split into:

- A set of 'comment format' definitions, e.g. prefix, start symbol, end symbol, etc. Should be able to support any language and any comment format, including block comments.
- A parser to read any comment into a tree of it's important parts. I tried to follow XML Parser's structure of nodes and node types here. Parser should be completely decoupled from anything language or version specific, and just parse a comment string into a tree.
- A writer to write back a comment tree into a proper formatted comment string using any of the 'format definitions'. Once again completely decoupled from the environment.
- Logic for interaction with either the `TextDocument` for legacy, or the Roslyn code model for modern visual studio versions. Roslyn already finds all the comments for us, this saves a lot of work.

All in all every time I got lost in trying to do too much at once, meaning a lot was done and nothing changed. I think ticket #63 would be a good start point to actually move the decisions for formatting into the writer and remove them from the parser. There is technically no difference between splitting the `<para>` and `<code>` tags on their own lines, and doing the same for a recursive list. Feel free to pick up the ticket and move it in that direction if you want.

willemduncan:
Since both this and #63 are still important functionality, I've addressed both points in 812654a30426f5f22a5c7dae74e7383b910fefd4, partially based on your code as well. Better to just fix a problem than to hold off until the day a rewrite appears. 

If you want, please check if the functionality of that version matches what you expect it to do, and if this PR would still be needed.

reima:
@willemduncan Thanks for moving forward with this issue. I've been too busy in the last weeks to take another look at it. I can confirm that your changes make this PR obsolete. Thanks again for picking it up!

willemduncan:
No problem, thanks for taking the time to look into it!

PR #172, opened 2015-10-24
Do not add XML indenting to literal content of nodes.

### LINKED ISSUE #141: Some comment tags are adversely affected by format comments
The first one I've seen that is no good is when using the `<see>` tag, which is often used inline in comment text, splitting tags to multiple lines affects these tags.  They should be ignored and allowed to remain inline if there is other text on the same line as the `<see>` tag.

The other issue I'm seeing is in regards to the `<code>` tag.  Here is how it looks before cleaning up comments:
```
        /// <summary>
        ///     Get the total memory installed on the Server
        ///     (in GB)
        /// </summary>
        /// <returns>
        ///     Total memory as <see cref="System.UInt32" />
        /// </returns>
        /// <example>
        ///     Method usage: 
        ///     <code>
        ///         using libServerInfo.Collectors;
        ///         using System.Windows;
        ///         ...
        ///         OS myOS = new OS();
        ///         
        ///         MessageBox.Show("Total RAM:  " + totalRam);
        ///         
        ///         myOS = null;
        ///     </code>
        /// </example>
```
Here is how it looks after cleaning up comments:
```
        /// <summary>
        ///     Get the total memory installed on the Server
        ///     (in GB)
        /// </summary>
        /// <returns>
        ///     Total memory as <see cref="System.UInt32" />
        /// </returns>
        /// <example>
        ///     Method usage: 
        ///     <code>
        ///                 using libServerInfo.Collectors;
        ///                 using System.Windows;
        ///                 ...
        ///                 OS myOS = new OS();
        ///         
        ///                 MessageBox.Show("Total RAM:  " + totalRam);
        ///         
        ///                 myOS = null;
        ///         </code>
        /// </example>
```

Notice that everything below the `<code>` tag up to and including the `</code>` tag are shifted right?  They will tab right once for each cleanup.  Now, since I do comment cleanup on a file save, I'm now have code samples that are shifted so far to the right they aren't even on screen anymore.  This makes me sad.

willemduncan:
Fixes  #141.


codecadwallader:
Thanks Willem!

PR #186, opened 2015-12-03
Comment formatting updates

willemduncan:
Adds a whole bunch of new unit tests to fix reported issues, and of course fixes the issues as well.

This fixes at least #141 (again), and on the way also fixed #166.


CodeMaid.UnitTests/Formatting/HeaderFormattingTests.cs
```diff
@@ -22,12 +22,6 @@ namespace SteveCadwallader.CodeMaid.UnitTests.Formatting
     [TestClass]
     public class HeaderFormattingTests
     {
-        [TestInitialize]
```
codecadwallader: I think by convention TestInitialize methods are usually at the top of a unit test file.  Any particular reason to move it to the bottom?  I'm guessing maybe our reorganization did it automatically and that should be a backlog card to make an exception for some unit test methods.

CodeMaid.UnitTests/Formatting/ListFormattingTests.cs
```diff
@@ -22,12 +22,6 @@ namespace SteveCadwallader.CodeMaid.UnitTests.Formatting
     [TestClass]
     public class ListFormattingTests
     {
-        [TestInitialize]
```
codecadwallader: Same as https://github.com/codecadwallader/codemaid/pull/186/files#r46758098

codecadwallader:
Overall looks very good to me, just the one question about the location of TestInitialize and then I can get this merged into master.

willemduncan:
Yes I was wondering about `[TestInitialize]` as well, they were moved down by the reorganize. I have left them there because I thought the CodeMaid project should dogfeed itself completely. Your solution specific settings even have `AutoCleanupOnFileSave = true`, and since I have "reorganize on cleanup" set, there was not much I could do. 

If you prefer, I will move them back to top, makes more sense to me as well.

codecadwallader:
Yep, I would prefer to move them back up top as that's the usual convention for unit tests. :)

I definitely agree about dogfooding CodeMaid and do that with both cleanup and comment formatting.  However I think of reorganization as a special case thing that I usually only run manually.  We do have the automatic options to run it because some people like to run it constantly, but it's off by default.  The reasoning of why I don't always run reorganization is if there's a logical order for members, I think that should supersede an alphabetical ordering.  A good example is our BaseCodeItem class, where the StartLine and StartOffset members comes before the EndLine and EndOffset members.

willemduncan:
Yes I'm one to always have 'auto reorganize on cleanup' set. Can we override that with the solution specific configuration?

codecadwallader:
Sure, want to go ahead and include it in this pull request?

willemduncan:
It should already be included automatically, at least I see the commit here and the changed files no longer include the `[TestInitialize]` changes.

codecadwallader:
Yep, the code looks good.  I was referencing if you wanted to check-in the CodeMaid.config with that setting change.  I'll make a note to change that soon and get this all merged in.  Thanks again! :)

willemduncan:
Added the config change as well.

codecadwallader:
Great, thanks!  Merged from my phone - how cool is that. :)

PR #193, opened 2015-12-16
Add undo transaction to insert region command.

reima:
Improves upon #182 by wrapping the operation in an undo transaction. Now the operation can be undone in one step (instead of around seven), similar to the remove region command.


codecadwallader:
Excellent addition - thank you!  Gladly accepted. :)

PR #198, opened 2016-01-12
Add support for VisualGDB's intellisence workaround

adontz:
VisualGDB reports *.c and *.cpp files' language as "C/C++ (VisualGDB)" and not "C/C++" if CLang intellisence is used.

### DIFF
--- a/CodeMaid/Helpers/CodeCommentHelper.cs
+++ b/CodeMaid/Helpers/CodeCommentHelper.cs
@@ -110,6 +110,7 @@ namespace SteveCadwallader.CodeMaid.Helpers
             switch (language)
             {
                 case "C/C++":
+                case "C/C++ (VisualGDB)":
                 case "CSharp":
                 case "CSS":
                 case "F#":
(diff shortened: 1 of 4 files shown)


codecadwallader:
Thanks for the pull request. :)  I'll take a look and give it consideration.

codecadwallader:
Did you test what happens when you don't have the document open and you right click on a file within the solution explorer?  When there isn't a document available, we instead utilize the ProjectItem that is available through the solution explorer to try to determine the language type.  The relevant code is here https://github.com/codecadwallader/codemaid/blob/master/CodeMaid/Logic/Cleaning/CodeCleanupAvailabilityLogic.cs#L304

I'm wondering if there may not be a distinct GUID for VisualGDB that we would need to account for as well.

adontz:
No, GUID is "{b2f072b0-abc1-11d0-9d62-00c04fd9dfd9}" so nothing to be done additionally.

codecadwallader:
Cool, thanks for confirming.  One other item - we have a special exception for end of file trailing new lines that does a specific check for C/C++ files.  I'm assuming the same workaround should apply with VisualGDB installed?

https://github.com/codecadwallader/codemaid/blob/master/CodeMaid/Logic/Cleaning/RemoveWhitespaceLogic.cs#L197-L199

adontz:
Fixed that too. Thanks for your review.

codecadwallader:
@adontz Thanks again for this pull request, it's been merged and will be present in the next beta release.  Do you mind if I credit you during the next release's blog post?

adontz:
I really do not think a few lines of code worth mentioning me, however it will be honor for me.

codecadwallader:
All contributions are appreciated! :)

Issue #215, opened 2016-02-12
Latest beta is losing the indentation of the first line during comment formatting


codecadwallader:
In the v0.9.0.1 BETA, I've noticed that when you run a comment formatting command (e.g. on an XML summary) the first line of the comment is losing its indentation and moving to the first column.

Before:
```csharp
    /// <summary>
    /// An implementation of <see cref="IConfigurationSectionHandler"/> that simply returns the provided XML.
    /// </summary>
```

After:
```csharp
/// <summary>
    /// An implementation of <see cref="IConfigurationSectionHandler"/> that simply returns the
    /// provided XML.
    /// </summary>
    ///
```

I'm not sure if this could be tied to changes made for #141?

If you have an opportunity @willemduncan can you please take a look?  If not I can dig in too.

codecadwallader:
I see some of the unit tests are failed as well:

- FormatStandardComments_FormatsAsExpected
- FormatStyleCopHeaderComments_FormatsAsExpected
- FormatXMLComments_FormatAsExpected

willemengel:
That is strange, all unit tests run correct on my fork, and there are no changes. Is this a problem just in the beta or does it also happen when you run the latest code?

willemduncan:
That is strange, all unit tests run correct on my fork, and there are no changes. Is this a problem just in the beta or does it also happen when you run the latest code?

codecadwallader:
That is strange, glad to hear you have a bead on it though - thanks! :)

codecadwallader:
As v0.9.1 is now released, closing this issue.

Issue #214, opened 2016-02-12
Issue when saving a C# file that has an empty comment


ChristineBoersen:
Given that I am a user,
When saving a C# file and Automatic Cleanup set to on and the file contains an empty single line comment (//)
Then the code needs to check for nulls properly as the following error is produced.

[CodeMaid Handled Exception 07:14:57 PM] Stopped cleaning 'DocumentClassificationGridController.cs': System.ArgumentNullException: Value cannot be null.
Parameter name: input
   at System.Text.RegularExpressions.Regex.Matches(String input)
   at SteveCadwallader.CodeMaid.Model.Comments.CommentFormatter.Parse(ICommentLine line, Int32 indentLevel, Int32 xmlTagLength)
   at SteveCadwallader.CodeMaid.Model.Comments.CommentFormatter..ctor(ICommentLine line, String commentPrefix, Int32 tabSize, Regex regex)
   at SteveCadwallader.CodeMaid.Model.Comments.CodeComment.Format()
   at SteveCadwallader.CodeMaid.Logic.Formatting.CommentFormatLogic.FormatComments(TextDocument textDocument, EditPoint start, EditPoint end)
   at SteveCadwallader.CodeMaid.Logic.Formatting.CommentFormatLogic.FormatComments(TextDocument textDocument)
   at SteveCadwallader.CodeMaid.Logic.Cleaning.CodeCleanupManager.RunCodeCleanupCSharp(Document document)
   at SteveCadwallader.CodeMaid.Logic.Cleaning.CodeCleanupManager.<>c__DisplayClass19_0.<Cleanup>b__0()
   at SteveCadwallader.CodeMaid.Helpers.UndoTransactionHelper.Run(Action tryAction, Action`1 catchAction)

codecadwallader:
Thanks for reporting the issue.  I tried to reproduce it but so far have been unsuccessful.  I attempted to reproduce it by having single line comments with `//`, files that were completely empty except for single line comments with `//`, as well as multi-line comments with single lines within them that had `//` but none of those seemed to trip the error condition you mentioned.

Some information that would be helpful:

- Visual Studio version
- CodeMaid version
- CodeMaid configuration (exportable through Options)
- Sample code file

ChristineBoersen:
Microsoft Visual Studio Professional 2015
Version 14.0.24720.00 Update 1
Microsoft .NET Framework
Version 4.6.01038

Installed Version: Professional

[...]

Issue #202, opened 2016-01-14
Run format comments during cleanup inserts trailing space


TheCycoONE:
Even when ```Remove end of line whitespace``` is enabled, ```Run format comments during cleanup``` causes a trailing space to be inserted on otherwise empty comment lines.

E.g. (using . to represent space)
```
///.<summary>
///.The following line is intentionally blank
///
///.Spiffy
///.</summary>
```
Becomes:
```
///.<summary>
///.The following line is intentionally blank
///.
///.Spiffy
///.</summary>
```
I would expect removing trailing white space to happen after and remove the trailing white space from the comment. This is needed to pass the default ```git diff --check```.

Using CodeMaid 0.9.0 on Visual Studio 2015. This is a C# project.

codecadwallader:
Thanks for reporting the issue.  I have been able to reproduce it as well just by manually invoking comment formatting (i.e. it doesn't have to be tied to a cleanup session).

The comment formatting logic is the very last step applied in the cleanup process, so that is why the end of line whitespace is not otherwise caught.

I'll leave it to @willemduncan to look into it further.  Thanks again!

kkm000:
I second that a comment formatting command adds trailing whitespace. Immediately noticeable with a trailing space highlighting plugin. :)

codecadwallader:
As v0.9.1 is now released, closing this issue.

PR #226, opened 2016-03-08
Fixes for #202, #214 and #215

willemduncan:
Included are a bunch of unit tests with comment prefixes to prevent regression. Not sure what happened to the `HeaderFormattingTests.cs` file, it's marked as completely changed even though it isn't. Not completely, anyway.


codecadwallader:
Awesome, thanks @willemduncan for the fixes!  I've confirmed the ones for 202 and 214.  The original issue for 215 seems fixed but I'm seeing something different now so I'll make a mention over there.

codecadwallader:
Apologies @willemduncan , the place where I was seeing a new issue was a solution with a lot of setting overrides for testing.  All three issues appear to be resolved in my testing. :)

willemduncan:
I did notice that the "local" cleanup settings impact the test results, might happen since the solution specific settings but I am not sure. Either way I'd say it's more of a defect on the testing side of things. 

Glad to hear the issues are fixed, anyway. I'm getting myself in an ever bigger knot trying to patch the current formatting code.

Issue #239, opened 2016-03-30
Reorganize moves variables out of #pragma warning blocks


bjorndavis:
### Environment
- Visual Studio version: 2013
- CodeMaid version: 0.9.1
- Code language: C#

### Description
When using the reorganize command, CodeMaid moves variables out of `#pragma warning` blocks.

### Steps to recreate
Before reorganize:
```
public class Test
    {
        private string test2;
#pragma warning disable 0649
        private string test1;
#pragma warning restore 0649
    }
```

After reorganize:
```
public class Test
    {
        private string test1;
        private string test2;

#pragma warning disable 0649
#pragma warning restore 0649
    }
```

### Current behavior
In certain specific cases it's desirable to disable compiler warnings for a section of code (e.g. disabling 0649 - "Field is never assigned to" when using the Selenium [PageFactory model.](https://github.com/SeleniumHQ/selenium/blob/eb0da4884fa118fa728d3b50ded08ff5355a9405/dotnet/src/support/PageObjects/PageFactory.cs#L80), as the variables are assigned using attributes).  However, CodeMaid will move variables out of these sections, which is undesirable.

### Expected behavior
Re-organize command should keep items within `#pragma warning` sections if they were there to begin with.

codecadwallader:
Thanks for reporting the issue, I have reproduced it.

When we detect other preprocessor conditionals (e.g. `#if`) we disable reorganization so it does not incorrectly manipulate the file.  We will now pick up `#pragma` statements as well during that detection and disable reorganization of files where that statement is present.

We have a new CI channel that will automatically generate builds based on the latest code snapshot.  If you would like to confirm the fix you can download it from here: http://vsixgallery.com/extension/4c82e17d-927e-42d2-8460-b473ac7df316/  Please note this is a CI channel so it is bleeding edge code and may exhibit other issues that the current released version does not.  If you encounter issues you can simply uninstall this version and re-install the latest stable version.

bjorndavis:
Disabling CodeMaid Reorganize entirely seems to be a fairly drastic step - is it possible to simply treat #pragma warning blocks like #regions (or at least enable/disable this in the settings)?  Unfortunately I'll have to stick to an older version as I have a fair amount of code that has to use the #pragmas, and I'll be unable to re-organize otherwise.

codecadwallader:
[...]

PR #269, opened 2016-04-08
Added safety check for StructLayout in CodeReorganizationAvailablityLogic

lukecummings:
Related to issue #239.

Required to prevent re-organization of structs or classes that are meant to be marshalled to a Native function (Order is important).

	[StructLayout(LayoutKind.Sequential)]
	public struct Example
	{
		int var1;
		int var3; //needs to stay above var2
		int var2;
	}

### DIFF
--- a/CodeMaid/Logic/Reorganizing/CodeReorganizationAvailabilityLogic.cs
+++ b/CodeMaid/Logic/Reorganizing/CodeReorganizationAvailabilityLogic.cs
@@ -64,7 +64,8 @@ namespace SteveCadwallader.CodeMaid.Logic.Reorganizing
                    document != null &&
                    document.GetCodeLanguage() == CodeLanguage.CSharp &&
                    !document.IsExternal() &&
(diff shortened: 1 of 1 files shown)


codecadwallader:
Thank you for submitting a pull request.  We actually already have logic in place to detect the StructLayout attribute and prevent reorganization from running.  It happens a little deeper down the stack which allows for you to reorganize a file but a particular member (e.g. a private struct within a class) is the only member that is skipped.

https://github.com/codecadwallader/codemaid/blob/master/CodeMaid/Logic/Reorganizing/CodeReorganizationManager.cs#L221

Are you seeing a scenario where this logic isn't working?

lukecummings:
Huh, well I'll be. I did have a case (the reason I went looking in the source in the first place) but now I can't seem to replicate it.

Sorry about that, my bad.

That'll teach me for submitting a pull request when I have a cold!

codecadwallader:
No worries.  If you stumble back into a scenario feel free to holler or submit a new pull. :)

PR #318, opened 2016-07-29
More Spade option

FlouFlou:
Hi codecadwallader,
Bravo for your work. 
I use only Spade because i was using VsCommands for VS 2010 File Structure. So i get used to see the structure of each file right the way it is coded (file order) for 3 years ! With public only properties, region, methods.... Recently i had to move to VS 2012 (i cant remember why) and the extension VsCommands for VS 2012 File Structure is totally useless (all properties are shown, no region...). So like i said it's very hard to change the way we are used to work. I need to see the file order structure because a lot of files in my solution are more than 2000 lines big. 
One day i felt on CodeMaid Spade and it was just amazing. The Spade window is almost the way i want.

I want to ask you some questions. 
- Is there a way to show only public properties (because i use a lot of 	mvvminpc properties) / methods / region) / region in methods ? (maybe with a !suffix / suffix or the key word private)
- Is there a way to Spade to keep the collapse/expand the way i leave it ? Because it's annoying when i do some code or switch to en other file and then Spade reload the entire structure and expand all.
- In this previous example, if i click to many times during the refresh, VS 2012 can freeze. I have to reboot the solution. I'm pretty sure it's CodeMaid fault because the Spade window seems to do some infinite loop or something and i think VS is waiting for it.

Thanks for your job and your reply


codecadwallader:
I do not think you meant to create a pull request, so I am closing it.

Issue #315, opened 2016-07-21
Properties that are explicit interface implementations are not sorted consistently


jpj625:
### Environment
- Visual Studio version: 2015 Enterprise Update 3
- CodeMaid version: 10.1
- Code language: C#
### Description

Properties with the same name, some of which are explicit interface implementations, are not sorted consistently. This causes occasional merge issues when the properties are re-ordered in different ways in different branches.
### Steps to recreate
1. Add attached file [ClassWithInconsistentSorting.txt](https://github.com/codecadwallader/codemaid/files/374941/ClassWithInconsistentSorting.txt) to a project (as .cs)
2. Apply Reorganize to the class ClassWithInconsistentSorting with "Alphabetize members of the same group" enabled
### Current behavior

Same-named properties (allowed by explicit interface implementation) sort differently based on initial position.

**Start**

```
IProperty Property
AreaB.IPropertyHaver.Property
AreaA.ICoProduct.Property
```

**Reorganize 1**

```
IProperty Property
AreaA.ICoProduct.Property
AreaB.IPropertyHaver.Property
```

**Reorganize 2**

```
AreaB.IPropertyHaver.Property
AreaA.ICoProduct.Property
IProperty Property
```

**Reorganize 3**

```
AreaA.ICoProduct.Property
AreaB.IPropertyHaver.Property
IProperty Property
```
### Expected behavior

Same-named properties should sort in a single, stable order.
Suggestion: Properties that are implicit implementations should be sorted first, with explicit implementations after, sorted by interface full name or other stable value.

codecadwallader:
Thanks for reporting the issue, I have reproduced it.  I agree the reorganization order should be stable and not require multiple runs to get to that state.

samcragg:
Hope I'm not hijacking with another feature request, but when implementing a fix would it be possible to have the option to have explicit interfaces methods/properties positioned after all public methods/properties - it might not be everyone's taste so that's why it would be good to have it optional?

codecadwallader:
@samcragg Do you know if there is any standard for how explicit interface implementations should be treated?  They are a bit of an oddball in that they don't have a declared access level but infer from their interface (if I remember correctly).  I took a quick peek at StyleCop rules but didn't see anything listed.

samcragg:
[...]

PR #337, opened 2016-09-13
Explicit interface order

samcragg:
Here's my attempt at fixing issue #315 and adding the ability to place explicit interface members after normal public methods.

Note I've added a dependency on NSubstitute (via NuGet) for the unit tests - I'm happy to change that to another mocking framework if you have a preference.

### DIFF
--- a/CodeMaid/Helpers/CodeItemTypeComparer.cs
+++ b/CodeMaid/Helpers/CodeItemTypeComparer.cs
@@ -70,6 +70,7 @@ namespace SteveCadwallader.CodeMaid.Helpers
         {
             int typeOffset = CalculateTypeOffset(codeItem);
             int accessOffset = CalculateAccessOffset(codeItem);
+            int explicitOffset = CalculateExplicitInterfaceOffset(codeItem);
             int constantOffset = CalculateConstantOffset(codeItem);
             int staticOffset = CalculateStaticOffset(codeItem);
             int readOnlyOffset = CalculateReadOnlyOffset(codeItem);
@@ -78,16 +79,16 @@ namespace SteveCadwallader.CodeMaid.Helpers
 
             if (!Settings.Default.Reorganizing_PrimaryOrderByAccessLevel)
             {
-                calc += typeOffset * 10000;
-                calc += accessOffset * 1000;
+                calc += typeOffset * 100000;
+                calc += accessOffset * 10000;
             }
             else
             {
-                calc += accessOffset * 10000;
-                calc += typeOffset * 1000;
+                calc += accessOffset * 100000;
+                calc += typeOffset * 10000;
             }
 
-            calc += (constantOffset * 100) + (staticOffset * 10) + readOnlyOffset;
+            calc += (explicitOffset * 1000) + (constantOffset * 100) + (staticOffset * 10) + readOnlyOffset;
 
             return calc;
         }
@@ -135,6 +136,20 @@ namespace SteveCadwallader.CodeMaid.Helpers
             if (codeItemField == null) return 0;
 
             return codeItemField.IsConstant ? 0 : 1;
+        }
+
+        private static int CalculateExplicitInterfaceOffset(BaseCodeItem codeItem)
+        {
+            if (Settings.Default.Reorganizing_ExplicitMembersAtEnd)
+            {
+                var interfaceItem = codeItem as IInterfaceItem;
+                if ((interfaceItem != null) && interfaceItem.IsExplicitInterfaceImplementation)
+                {
+                    return 1;
+                }
+            }
+
+            return 0;
         }
 
         private static int CalculateStaticOffset(BaseCodeItem codeItem)
--- a/CodeMaid/Helpers/CodeItemTypeComparer.cs
+++ b/CodeMaid/Helpers/CodeItemTypeComparer.cs
@@ -1,7 +1,7 @@
-using System.Collections.Generic;
 using EnvDTE;
 using SteveCadwallader.CodeMaid.Model.CodeItems;
 using SteveCadwallader.CodeMaid.Properties;
+using System.Collections.Generic;
 
 namespace SteveCadwallader.CodeMaid.Helpers
 {
@@ -130,14 +130,6 @@ namespace SteveCadwallader.CodeMaid.Helpers
             }
         }
 
-        private static int CalculateConstantOffset(BaseCodeItem codeItem)
-        {
-            var codeItemField = codeItem as CodeItemField;
-            if (codeItemField == null) return 0;
-
-            return codeItemField.IsConstant ? 0 : 1;
-        }
-
         private static int CalculateExplicitInterfaceOffset(BaseCodeItem codeItem)
         {
             if (Settings.Default.Reorganizing_ExplicitMembersAtEnd)
@@ -150,6 +142,14 @@ namespace SteveCadwallader.CodeMaid.Helpers
             }
 
             return 0;
+        }
+
+        private static int CalculateConstantOffset(BaseCodeItem codeItem)
+        {
+            var codeItemField = codeItem as CodeItemField;
+            if (codeItemField == null) return 0;
+
+            return codeItemField.IsConstant ? 0 : 1;
         }
 
         private static int CalculateStaticOffset(BaseCodeItem codeItem)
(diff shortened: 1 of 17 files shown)


codecadwallader:
Thanks so much for putting this together.  I'd like to look over it a little more, test it out and I'm sure I'll have a few comments - but at first blush it looks really good.

I haven't used NSubstitute before.. but I've used Moq and Microsoft Fakes which are similar.  NSubstitute looks pretty straight-forward to me though so seems like we should roll with it.  As you may have noticed I've done very little mocking, and very heavy integration testing through the IDE.  It's more realistic, but it's also extremely slow and not CI friendly - so glad to see more pure unit tests.

samcragg:
Thanks for the kind comments - there's no rush as far as I'm concerned; the code in the project is in a real good shape to expand on (I was dreading figuring out how to add a setting but when I looked at the code it was very simple).

Like I said, I can swap out NSubstitute for Moq if you're more familiar with it - it's only used in one place to mock the CodeFunction2 interface, as it had far too many members to create a manual fake I figured it was easier to just use a mocking framework.

BTW, it looks like the CI server can't restore NuGet packages - since it's only the unit test project I figured it wouldn't hurt brining in an outside dependency but I'm not sure how to get the CI to do a nuget restore on it, sorry.

codecadwallader:
I'm glad you found the code accessible. :)  Having something like 160 settings already does make that a pretty well-worn pattern. ;)

No need to swap out NSubstitute - again it looks pretty straight-forward and much lighter than Microsoft Fakes as an example.

I'll take a look into the CI server failing the builds.  We are using NuGet for the main CodeMaid project so it already knows how to restore packages, but maybe it isn't doing it across the solution.

CodeMaid/Helpers/CodeItemTypeComparer.cs
```diff
@@ -1,7 +1,7 @@
+using System.Collections.Generic;
```
codecadwallader: Nitpick, but can you sort the using statements alphabetically?  This is the default as of VS2012, even though previously System always came first and some tools like ReSharper still enforce it.

CodeMaid/Helpers/CodeItemTypeComparer.cs
```diff
@@ -123,6 +138,20 @@ private static int CalculateConstantOffset(BaseCodeItem codeItem)
             return codeItemField.IsConstant ? 0 : 1;
         }
 
+        private static int CalculateExplicitInterfaceOffset(BaseCodeItem codeItem)
```
codecadwallader: Another nitpick, but can you move this above CalculateConstantOffset so these methods appear in the order they are invoked?

CodeMaid/Helpers/CodeItemTypeComparer.cs
```diff
@@ -139,6 +168,23 @@ private static int CalculateReadOnlyOffset(BaseCodeItem codeItem)
...
 
+        private static string NormalizeName(BaseCodeItem codeItem)
+        {
+            string name = codeItem.Name;
+            var interfaceItem = codeItem as IInterfaceItem;
+            if ((interfaceItem != null) && interfaceItem.IsExplicitInterfaceImplementation)
+            {
+                // Try to find where the interface ends and the method starts
+                int dot = name.LastIndexOf('.') + 1;
+                if ((dot > 0) && (dot < name.Length))
```
codecadwallader: A convention I've started using for 'between' checks is to put the variable in the middle, with the smaller check on the left and the larger check on the right.  Minor readability enhancement.  So for example:

`if (0 < dot && dot < name.Length)`

CodeMaid/Helpers/CodeItemTypeComparer.cs
```diff
@@ -139,6 +168,23 @@ private static int CalculateReadOnlyOffset(BaseCodeItem codeItem)
             return codeItemField.IsReadOnly ? 0 : 1;
         }
 
+        private static string NormalizeName(BaseCodeItem codeItem)
+        {
+            string name = codeItem.Name;
+            var interfaceItem = codeItem as IInterfaceItem;
+            if ((interfaceItem != null) && interfaceItem.IsExplicitInterfaceImplementation)
```
codecadwallader: This could be re-written as `if (interfaceItem?.IsExplicitInterfaceImplementation ?? false)`
Very debatable which is easier to read, so just food for thought. :)

CodeMaid/Logic/Reorganizing/CodeReorganizationManager.cs
```diff
@@ -364,9 +364,8 @@ private void RecursivelyReorganize(IEnumerable<BaseCodeItem> codeItems, ICodeIte
 
             // Get the items in their current order and their desired order.
             var currentOrder = GetReorganizableCodeItemElements(codeItems);
-            var desiredOrder = currentOrder.OrderBy(CodeItemTypeComparer.CalculateNumericRepresentation)
-                                           .ThenBy(x => Settings.Default.Reorganizing_AlphabetizeMembersOfTheSameGroup ? (object)x.Name : (object)x.StartOffset)
-                                           .ToList();
+            var desiredOrder = new List<BaseCodeItemElement>(currentOrder);
+            desiredOrder.Sort(new CodeItemTypeComparer(Settings.Default.Reorganizing_AlphabetizeMembersOfTheSameGroup));
```
codecadwallader: This certainly looks simpler then two separate sort passes.  Are you convinced it has the same effect?

codecadwallader:
Tada, build now passing. :)

CodeMaid/Helpers/CodeItemTypeComparer.cs
samcragg: Initialially I wrote it as `if (interfaceItem?.IsExplicitInterfaceImplementation == true)` but didn't like the lifting to nullable and also it's (for me!) weird to have if statements explicitly comparing for true, as it looks redundant.

CodeMaid/Helpers/CodeItemTypeComparer.cs
samcragg: I do normally as well - this one slipped by as it evolved into that (initially it was just a check for -1, but then I added to the index and then I thought I best be safe about index out of bounds)

CodeMaid/Logic/Reorganizing/CodeReorganizationManager.cs
samcragg: I'm quite confident. Although List.Sort isn't a stable sort, I don't think we'll ever return two things as equal as we use StartOffset as a tiebreaker (and if that was the same it wouldn't matter switching them around ;-) )

The ThenBy part is handling inside the comparer, as it was already there for the Spade tool. Since they use different settings the flag to switch on name sorting is passed into the constructor.

CodeMaid/Helpers/CodeItemTypeComparer.cs
codecadwallader: Makes sense and agree, thanks.

CodeMaid/Logic/Reorganizing/CodeReorganizationManager.cs
codecadwallader: Cool, thank you for confirming.  Agree that StartOffset should break any ties, as even variables on the same line (e.g. `int x, y;`) have different character offsets.

codecadwallader:
Thanks for the quick fixes.  I'll play around with this a little locally with the plan to merge this weekend.  After merging it will be auto-deployed to the [CI channel](http://vsixgallery.com/extension/4c82e17d-927e-42d2-8460-b473ac7df316/) so there will be a build you can install for use as well as it getting pushed out to anybody setup with a direct feed to that channel.

codecadwallader:
Ran it through a couple paces and it worked well for me.  Merging! :)  As mentioned above, feel free to confirm it in the CI channel build.  Thanks again for contributing!  I hope you don't mind being accredited in the next release notes?

samcragg:
Thanks for that, just installed the CI and ran it over one of my projects that has some explicit interface methods mixed with public ones and it handles all the files now perfectly (for my OCD anyway :D )

I have no objection to my name appearing anywhere, though it's not necessary as I just wanted to give back to something I've used for years and love (it's the first thing that gets installed in VS).

Keep up the great work!

codecadwallader:
Awesome, that's really nice and rewarding to hear. :)

Issue #326, opened 2016-08-16
CodeMaid Spade: Visual Basic comments are not displayed


aeab13:
### Environment
- Visual Studio version: 2015 Community
- CodeMaid version: 10.1
- Code language: VB .NET
### Description

Documentation comments are not displayed in CodeMaid Spade using VB .NET but they work using C#. Is this an issue?
### Steps to recreate
1. Create an example method with documentation comments.
2. Check CodeMaid Spade. 
   ![2016-08-16 16_56_13-consoleapplication5 - microsoft visual studio](https://cloud.githubusercontent.com/assets/16917468/17703556/6b1cf320-63d2-11e6-8729-bd76a9d06c82.png)
   ![2016-08-16 16_56_27-consoleapplication1 - microsoft visual studio](https://cloud.githubusercontent.com/assets/16917468/17703557/6b3d2208-63d2-11e6-97a8-641f1905837f.png)
### Current behavior

Comments are not displayed.
### Expected behavior

Comments should be displayed.

codecadwallader:
Thanks for reporting the issue, I have reproduced it.  I found a minor discrepancy in how VB and C# report the XML hierarchy that was causing us not to present the comment in the case of VB.  There was a very simple fix for it, which I've committed and will be available in our [CI channel](http://vsixgallery.com/extension/4c82e17d-927e-42d2-8460-b473ac7df316/) in a few moments (10.1.100 or higher) if you would like to test it out.  This will be a part of the next minor release as well.

Thanks again!

aeab13:
Now it is working! Thanks!
![2016-08-18 09_41_13-consoleapplication5 - microsoft visual studio](https://cloud.githubusercontent.com/assets/16917468/17765737/07775188-6528-11e6-85c0-ca62a490d51f.png)
![Uploading 2016-08-18 09_41_24-ConsoleApplication6 - Microsoft Visual Studio.png…]()

aeab13:
Now it is working! Thanks for fixing it!
![2016-08-18 09_41_13-consoleapplication5 - microsoft visual studio](https://cloud.githubusercontent.com/assets/16917468/17765756/1d1e577a-6528-11e6-909b-52e92d10bf05.png)
![2016-08-18 09_41_24-consoleapplication6 - microsoft visual studio](https://cloud.githubusercontent.com/assets/16917468/17765757/1d23197c-6528-11e6-98c9-c12321d0dadd.png)

codecadwallader:
[...]

PR #341, opened 2016-09-17
Now adds <doc> root tag if it doesn't exist

aeab13:
This change tries to fix the issue [#326](https://github.com/codecadwallader/codemaid/issues/326).
I've added a condition that checks that the `<doc>` tag exists and adds it if does not exist.


codecadwallader:
Thank you!  This is now deploying directly to the CI channel: http://vsixgallery.com/extension/4c82e17d-927e-42d2-8460-b473ac7df316/

PR #345, opened 2016-09-23
Fix for issue 329: Cleanup removes consecutive blank lines in a multiline string literal.

blavigne13:
I just tweaked the regex for 3+ newlines to do a look-behind assertion in order to ensure that it is not inside of an un-terminated verbatim string literal.

### DIFF
--- a/CodeMaid/Logic/Cleaning/RemoveWhitespaceLogic.cs
+++ b/CodeMaid/Logic/Cleaning/RemoveWhitespaceLogic.cs
@@ -218,7 +218,7 @@ namespace SteveCadwallader.CodeMaid.Logic.Cleaning
         {
             if (!Settings.Default.Cleaning_RemoveMultipleConsecutiveBlankLines) return;
 
-            const string pattern = @"(\r?\n){3,}";
+            const string pattern = @"(?<!@""(""""|[^""])*)(\r?\n){3,}";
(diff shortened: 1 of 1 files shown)


codecadwallader:
Thanks for putting this together.  Logically it makes sense, and I've run a few tests trying to break it and it has worked as expected.

My only remaining concern with it is the added performance impact from performing the negative  lookbehind checks.  I'm not sure how far back in the document it would have to scan to disprove that it is a string literal.  I'm definitely not a regular expression expert, but since any line not containing a pair of double quotes qualifies for the latter part of the expression, I would think that this would usually have to scan each time all the way to the top of the document to "pass" the negative lookbehind section.  This wouldn't matter much in shorter files, but certainly in longer ones it could start to get expensive.

Any interest in doing some deeper analysis/testing to see how it performs with large files (10,000 is our high end benchmark) where there are multiple newlines near the end and/or figuring out how to confirm if it does indeed scan all the way back to the top of the document every time?

blavigne13:
Sorry for taking so long to respond, I was out of town for a bit. I actually had the same concern about the look-behind. Since it's linear, I don't expect it to be a problem, but I'll throw together a bunch of large text files and see how it performs. I doubt I'll get to it tonight, but probably some time this week.

PR #359, opened 2016-10-25
Order by access modifier now configurable.

ahalassy:


CodeMaid/Helpers/CodeItemTypeComparer.cs
```diff
@@ -118,16 +119,20 @@ private static int CalculateAccessOffset(BaseCodeItem codeItem)
...
-            {
-                case vsCMAccess.vsCMAccessPublic: return 1;
-                case vsCMAccess.vsCMAccessAssemblyOrFamily: return 2;
-                case vsCMAccess.vsCMAccessProject: return 3;
-                case vsCMAccess.vsCMAccessProjectOrProtected: return 4;
-                case vsCMAccess.vsCMAccessProtected: return 5;
-                case vsCMAccess.vsCMAccessPrivate: return 6;
-                default: return 0;
-            }
+            var itemsOrder = new[] {
```
codecadwallader: Instead of starting in arrays, and converting to list.. why not just use lists?

CodeMaid/Helpers/CodeItemTypeComparer.cs
```diff
@@ -118,16 +119,20 @@ private static int CalculateAccessOffset(BaseCodeItem codeItem)
...
+                vsCMAccess.vsCMAccessPublic,
+                vsCMAccess.vsCMAccessAssemblyOrFamily,
+                vsCMAccess.vsCMAccessProject,
+                vsCMAccess.vsCMAccessProjectOrProtected,
+                vsCMAccess.vsCMAccessProtected,
+                vsCMAccess.vsCMAccessPrivate
+                };
+            if (!Settings.Default.Reorganizing_RevertAccessLevel)
```
codecadwallader: Revert isn't a clear verb to me (means go back to an original state).  I think you mean Reverse access level.  In that case though, wouldn't this be a positive case where only if the flag was set would the reversal happen?

CodeMaid/Helpers/CodeItemTypeComparer.cs
```diff
@@ -118,16 +119,20 @@ private static int CalculateAccessOffset(BaseCodeItem codeItem)
...
+                itemsOrder = itemsOrder.Reverse().ToArray();
+
+            return itemsOrder.Contains(codeItemElement.Access)
```
codecadwallader: IndexOf always returns -1 when no match is found, so you could remove the conditional and always return IndexOf+1.

CodeMaid/UI/Dialogs/Options/Reorganizing/ReorganizingGeneralDataTemplate.xaml
```diff
@@ -15,6 +15,18 @@
                         <RadioButton Content="type then access" IsChecked="{Binding PrimaryOrderByAccessLevel, Converter={x:Static cnv:BooleanInverseConverter.Default}}" />
                         <RadioButton Content="access then type" IsChecked="{Binding PrimaryOrderByAccessLevel}" />
                     </StackPanel>
+                    <StackPanel
```
codecadwallader: More nitpicks:

* Can we follow the formatting conventions of the rest of the file (e.g. attributes on same line)
* Can we use the same verbiage as other options.. (e.g. "Access levels should be ordered by" with the default option being first "public to private" and the alternate option being second "private to public")

CodeMaid/UI/Dialogs/Options/Reorganizing/ReorganizingGeneralViewModel.cs
```diff
@@ -25,7 +25,8 @@ public ReorganizingGeneralViewModel(CodeMaidPackage package, Settings activeSett
                 new SettingToOptionMapping<bool, bool>(x => ActiveSettings.Reorganizing_KeepMembersWithinRegions, x => KeepMembersWithinRegions),
                 new SettingToOptionMapping<int, AskYesNo>(x => ActiveSettings.Reorganizing_PerformWhenPreprocessorConditionals, x => PerformWhenPreprocessorConditionals),
                 new SettingToOptionMapping<bool, bool>(x => ActiveSettings.Reorganizing_PrimaryOrderByAccessLevel, x => PrimaryOrderByAccessLevel),
-                new SettingToOptionMapping<bool, bool>(x => ActiveSettings.Reorganizing_RunAtStartOfCleanup, x => RunAtStartOfCleanup)
+                new SettingToOptionMapping<bool, bool>(x => ActiveSettings.Reorganizing_RunAtStartOfCleanup, x => RunAtStartOfCleanup),
+                new SettingToOptionMapping<bool, bool>(x => ActiveSettings.Reorganizing_RevertAccessLevel, x => RevertOrderByAccessLevel)
```
codecadwallader: Will you please move this up to continue the alphabetical ordering.

CodeMaid/UI/Dialogs/Options/Reorganizing/ReorganizingGeneralViewModel.cs
```diff
@@ -96,6 +97,18 @@ public bool RunAtStartOfCleanup
             set { SetPropertyValue(value); }
         }
 
+        /// <summary>
+        /// Gets or sets a value indicating whether [revert order by access level].
+        /// </summary>
+        /// <value>
+        /// <c>true</c> if [revert order by access level]; otherwise, <c>false</c>.
+        /// </value>
+        public bool RevertOrderByAccessLevel
```
codecadwallader: Will you please move this up to continue the alphabetical ordering.  It would also be helpful to match the comment style to the others vs. auto-generated.

Perhaps a clearer name for this property would be ReverseAccessLevelOrdering

ahalassy:
Thanks for all the notes I am going to fix the code keeping this in my mind. Also, I noticed that the code ordering puts the static members to the end of the code however I need the opposite ordering in my codes. Anyway, I also implemented configuration for for static member ordering in my fork. I am going to review that taking in consideration your notes and add those changes to this pull request within a few days. Do you agree or should I handle this on a different request?

codecadwallader:
Sounds good.  It's pretty closely related so I'm OK with it being on the same pull request.  You may already be thinking it but it sounds like a separate configuration option to reverse static members.

ahalassy:
It definitely is a separate configuration option with it's own GUI option and setting field.

ahalassy:
Hi! Sorry for the long delay, I was quite busy during the last few months. I made several changes to align with your conventions. Can you have a look?

codecadwallader:
Looks good to me!  Merging in to master now. :)  Thanks again for your contribution!  I hope you don't mind if I credit you during the next release?

codecadwallader:
I made a couple formatting tweaks.. and I'm questioning the reversal logic now.  The setting is off by default (i.e. public -> private).. yet in CodeItemTypeComparer it is reversing the items order when the setting is not true (i.e. off).  Should the exclamation mark be removed, or am I looked at it backwards?

ahalassy:
I need to check this. Previously I planned to create a similar enhancement for static-instance level ordering - I am going to implement this. I'll check this as well, and I'll inform you if further alignments needed. My original intention was to keep your logic as default.

codecadwallader:
Ok, sounds good.  I went ahead and flipped that statement as I'm pretty sure from tracing through the logic it was inverted.  Thanks for taking a look!

codecadwallader:
I did retest that logic and confirm it's correct now.

PR #384, opened 2017-01-06
Remove C# Alias to NET CLR

robertodalmonte:
I like String instead of string, Boolean instead of bool etc...
I actually use AliasBeGone extension
https://marketplace.visualstudio.com/items?itemName=PerBrage.AliasBeGone

I'd love to see this functionality within CodeMaid.
Regards
Roberto Dalmonte


codecadwallader:
It seems you have created this as a pull request, I think you meant to create an issue.

If the functionality already exists in another extension, why re-create it here?

Issue #393, opened 2017-01-26
FeatureRequest: Clear Solution Explorer search query before Find in Solution Explorer


joeburdick:
### Environment
- Visual Studio version: 2015
- CodeMaid version: 10.2
- Code language: N/A

### Description
Currently, if you have a search query in your Solution Explorer and then use the "Find in Solution Explorer" feature on a file which does not match your search query, it instead opens a file which is currently in the Solution Explorer.

### Steps to recreate
1. Open a file.
2. Enter a search query into Solution Explorer which will not have your open file as a result.
3. Use "Find in Solution Explorer" on your open file.

### Current behavior
A file in your filtered Solution Explorer is opened.

### Expected behavior
The Solution Explorer's search should be cleared, and then the currently open file should be found in Solution Explorer.

Maybe this can be an option in settings?

codecadwallader:
Thanks for the suggestion.

A code starting point for anyone interested in pursuing: https://github.com/codecadwallader/codemaid/blob/master/CodeMaid/Integration/Commands/FindInSolutionExplorerCommand.cs#L53

codecadwallader:
Thanks for implementing this!

PR #394, opened 2017-01-30
#393: Add ability to clear solution explorer search before finding.

joeburdick:
This solution works but has one potential drawback. Though the Solution Explorer will stop filtering items by the search query, the query string does not leave the search box. I could not find a way to make this box clear (and I am not even sure that is the behavior one would want.)

### DIFF
--- a/CodeMaid/CodeMaidPackage.cs
+++ b/CodeMaid/CodeMaidPackage.cs
@@ -22,6 +22,7 @@ using System.Linq;
 using System.Runtime.InteropServices;
 using System.Windows;
 using System.Windows.Threading;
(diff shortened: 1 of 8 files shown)


codecadwallader:
Thanks for putting this together!  I pushed some minor tweaks, including promoting when the IServiceProvider is initialized to resolve a unit test execution issue.

I think the text still being present is a little awkward, but since it's an opt-in feature it's something that can be improved upon.  Merging in accordingly.

codecadwallader:
The CI build with this change can be found here for testing: http://vsixgallery.com/extension/4c82e17d-927e-42d2-8460-b473ac7df316/

Issue #333, opened 2016-09-08
VB: Drag & Drop Re-ordering does not move attributes.


ericmutta:
### Environment
- Visual Studio 2015 Community Update 3
- CodeMaid 10.1
- VB,NET
### Description

In the CodeMaid Spade window you can drag and drop members up or down to re-organize them physically within a file. This works great but if the methods contain an attribute, the method body moves by the attributes applied to the method are left behind.
### Steps to recreate

Try to re-organise the following code using CodeMaid Spade (e.g. by moving BlockWrite() up):

`Public Module ArrayExtensionMethods
  &lt;Extension>
  Public Function BlockRead(ArgSource As Byte(), ArgOffset As Integer, ArgCount As Integer) As Byte()
    Dim bytes = New Byte(ArgCount - 1) {}
    Buffer.BlockCopy(ArgSource, ArgOffset, bytes, 0, bytes.Length)
    Return bytes
  End Function

  &lt;Extension>
  Public Sub BlockWrite(ArgSource As Byte(), ArgTarget As Byte(), ArgOffset As Integer)
    Buffer.BlockCopy(ArgSource, 0, ArgTarget, ArgOffset, ArgSource.Length)
  End Sub
End Module`
### Current behavior

The method bodies are moved by the applied Extension attribute gets left standing alone which results in a syntax error.
### Expected behavior

Move the method including its applied attributes. For extra value, the surrounding blank lines should be preserved after move so if someone uses blank lines to seperate methods, they apply after the move.

codecadwallader:
Thanks for reporting the issue.  This behavior works as expected in C#, so I suspect that the VB API is not reporting the StartPoint member for CodeElement's in the same way that occurs in C#.  Another way to test this behavior would be to perform a reorganization operation and see if the same things happens.  Under the covers, it's largely the same logic that is used to move elements within a file between Spade's drag and drop and the reorganization logic.

The code that determines the starting point of elements: https://github.com/codecadwallader/codemaid/blob/master/CodeMaid/Model/CodeItems/BaseCodeItemElement.cs#L67

In regards to the surrounding blank lines - it is expected that cleanup will be run after any move/reorganization operations which will re-standardize the blank lines.

ericmutta:
@codecadwallader many thanks for getting back to me and confirming the bug. I hope it will be fixed in a future release (I would help, but have no experience with Visual Studio plug-in development!).

[...]

PR #444, opened 2017-05-05
VB.Net Code Element Positions: Fixing issue #333

thehutman:
Fix for #333 VB: Drag & Drop Re-ordering does not move attributes.

Fundamentally, the `CodeElement.StartPoint` works differently in VB.Net than it does for C#.  In VB.Net, the property does NOT include attributes.  It does for C#.  Be nice if VB.Net and C# played by the same rules, but here we are.

In order to make this work the same in both versions, I modified code to use the `GetStartPoint` method, rather than the property and uses the `vsCMPart.vsCMPartWholeWithAttributes` flag to get the entire code segment with attributes.  Also did the same for `EndPoint` (although I don't think there were problems with it, I did it to be consistent).

Things working "wrong" in VB.Net before this change included:
- Reorganization.  Items with attributes were corrupted.
- Dragging around items in Code Spade.  Items with attributes were corrupted.
- Double-clicking items in Code Spade to highlight the entire code block.  Items with attributes did not select the attribute (and any comments before the attribute block)

With this change VB.Net now behaves the same as C# (as much as I am able to tell from testing against a variety of source code I am using).  C# doesn't appear to be behaving any differently than before.

I know this may be a low level change, but I believe this change mimics what was originally intended (the `StartPoint`/`EndPoint` pair needs to include the attributes).


thehutman:
I've also added some updates to handle formatting of VB.Net code (line spacing in particular).  The original CodeCleanup mehtods were using a very minimal `RunCodeCleanupGeneric` for VB.Net.  I added a new `RunCodeCleanupVB` that has a good portion of what was in `RunCodeCleanupCSharp` which now formats all of the spacing as configured.

codecadwallader:
Thank you, I sincerely appreciate the time you have put into these changes and think this is well under way to being accepted. :)

I've been reviewing and testing the changes this morning.  I confirmed what you saw, there was no difference in C# between the StartPoint property and the GetStartPoint() method but there was a difference in VB.  With that knowledge I was more comfortable with the location of the change.  I made a couple very minor edits and pushed them into your branch.

I also reviewed many (hopefully all) other occurrences of StartPoint/EndPoint outside of BaseCodeItemElement and they were logically fine the way they were.  For example calculating complexity only uses the internal component of the method and the attributes are not relevant.

For the second stage of your changes (extending VB cleanups) its a great extension of VB capabilities.  How much have you tested each component of those changes to make sure they are VB friendly?  For example I believe UpdateEndRegionDirectives is only written to support C# region directives.  It would be very helpful to test each component of that cleanup (like you have with explicit access modifiers) to make sure they are VB compatible.  Also as a general convention I would ask that commented out code be deleted.

Thanks again!

thehutman:
You are welcome!  Oh the fun to be had waiting on laundry :)

I'm not a heavy user of regions.  I did some basic tests and they seemed to act accordingly in the VB code I have.  I did test the formatting pretty extensively.  This is how I found I needed to remove some of the access modifier stuff since it was breaking Async methods.  I believe this is because the access modifiers aren't the same in VB as they are for C# and the modifier text is hard coded to the C# language.  So I commented it out for now.

I removed the commented code as asked :)  I was hesitant on if I really wanted it removed or if I might code up the workaround.  As it is now, I think its "good enough" as it works well now in our rather large VB.Net solution.

codecadwallader:
Awesome, thanks for the tweaks.  I did one more sanity pass through the code and removed the UpdateEndRegionDirectives call.  It wasn't causing any harm since it was looking for C# directive strings to match on, but it also wasn't doing anything productive.

Pull request accepted, merged, and now available in our CI channel here for testing :) http://vsixgallery.com/extension/4c82e17d-927e-42d2-8460-b473ac7df316/

This is a really welcome improvement and I'm very grateful for the time you took to do it.  I'll plan to push this out to the public as a v10.4 release in a couple weeks following some time for any reports to come in from users of the CI channel.  Thanks again!

PR #513, opened 2018-01-12
Add the ability to add variables to the automatically generated file header

Yannike:
Add the ability to add variable to the header like the filename, the project name, the solution name ...
Modify the clean update section in the update dialog to add variable to the header


codecadwallader:
Thank you for providing this pull request!  Can you provide an explanation of your approach?

Yannike:
By explanation, you mean how it works?

codecadwallader:
Yes please, it will help inform the review.

Yannike:
In the option dialog, you can add variables to the header added automatically to every files. I added buttons in the dialog to make adding the variables easier. So when a header needs to be added if it has $FILENAME$ (for example), this gets replaced by the name of the current file and so on for the rest of the available variables. 

This was done by adding the 'variable buttons' in the option dialog ( the clean/update subpage) and then slightly modifying the code when the header is added to check if there is any variable and to replace them by their actual value.

CodeMaid/Logic/Cleaning/FileHeaderLogic.cs
```diff
@@ -58,6 +61,8 @@ internal void UpdateFileHeader(TextDocument textDocument)
                 return;
             }
 
+            this.ReplaceVariables(textDocument, ref settingsFileHeader);
```
codecadwallader: Please remove the redundant 'this'.
Instead of using ref arguments, prefer to pass input/output parameters to the method.

CodeMaid/Logic/Cleaning/FileHeaderLogic.cs
```diff
@@ -68,6 +73,58 @@ internal void UpdateFileHeader(TextDocument textDocument)
             }
         }
 
+        /// <summary>
+        /// Replace the variables in the header if any
+        /// </summary>
+        /// <param name="textDocument"></param>
+        internal void ReplaceVariables(TextDocument textDocument, ref string settingsFileHeader)
+        {
+            foreach (var variable in _variables)
```
codecadwallader: This approach will be making calls out to the IDE every time it is invoked.. even if the header doesn't contain any of these variables.

I think perhaps instead it would make sense to scan the header for the variables, determine which were present, calculate the replacement and put it in place.

CodeMaid/UI/Dialogs/Options/Cleaning/CleaningUpdateViewModel.cs
```diff
@@ -227,5 +228,41 @@ public bool UpdateSingleLineMethods
...
+        {
+            var values = (object[])parameter;
+            if (values.Length != 2)
+            {
+                return;
+            }
+
+            var tabControl = values[0];
+            var variableText = values[1] as string;
+            var textBox = (((tabControl as System.Windows.Controls.TabControl)?.SelectedItem as System.Windows.Controls.TabItem)?.Content as System.Windows.Controls.TextBox);
```
codecadwallader: Having the view model have hard-coded view references within it breaks the intended flow of View->ViewModel.  It looks like this is just trying to inject some text placeholders.  This could be done by passing the text through the CommandParameter and binding that in the UI.  Then the ViewModel wouldn't need to understand TabControl's, SelectedItems, etc.

CodeMaid/UI/Dialogs/Options/Cleaning/CleaningUpdateDataTemplate.xaml
```diff
@@ -13,7 +14,7 @@
             </GroupBox>
             <GroupBox Header="File Header">
                 <StackPanel>
-                    <TabControl>
+                    <TabControl Name="LanguageTab">
```
codecadwallader: Based on comments below, the Name shouldn't be needed for ViewModel access.

CodeMaid/UI/Converters/AppendParametersConverter.cs
```diff
@@ -0,0 +1,44 @@
+﻿using System;
```
codecadwallader: I don't think this class is needed at all when the TabControl is taken out of the ViewModel (comments below).

CodeMaid/UI/Dialogs/Options/Cleaning/CleaningUpdateDataTemplate.xaml
```diff
@@ -63,6 +64,42 @@
                             <TextBox Style="{DynamicResource TextBoxMultiLine}" Text="{Binding UpdateFileHeaderTypeScript, UpdateSourceTrigger=PropertyChanged}" />
                         </TabItem>
                     </TabControl>
+                    <GroupBox Header="Variables">
+                        <StackPanel Orientation="Horizontal">
+                            <Button Name="UserLogin" Content="$USER_LOGIN$" Width="Auto" Margin="2, 2, 2, 2" Command="{Binding AddHeaderVariableCommand}">
+                                <Button.CommandParameter>
```
codecadwallader: I think these bindings will be simplified after removing the View code from the ViewModel.

codecadwallader (changes requested):
Ok, thanks for the explanation - comments below! :)

CodeMaid/UI/Dialogs/Options/Cleaning/CleaningUpdateViewModel.cs
Yannike: Okay, so I don't totally get how you want to do that. Here is what we need to have
1- textbox => we need to get the currentposition of the cursor to insert the text at the right place
2- text to add => the variable

How do you suggest binding everything together in the view to be able to just pass the text to add as parameter to the command ?

codecadwallader: What if you passed the text, the index (i.e. cursor position), and the variable name?

Yannike: This doesn't work as well. My idea was to have something like that :
```
<Button Name="Project" Content="$PROJECT$" Width="Auto" Margin="2, 2, 2, 2" Command="{Binding AddHeaderVariableCommand}">
    <Button.CommandParameter>
        <MultiBinding Converter="{x:Static cnv:AppendParametersConverter.Default}">
            <Binding ElementName="LanguageTab" Path="SelectedItem.Content.SelectionStart"/>
            <Binding ElementName="LanguageTab" Path="SelectedItem.Content.Text"/>
            <Binding ElementName="FileName" Path="Content" />
        </MultiBinding>
    </Button.CommandParameter>
</Button>
```

First, this still requires the elementName (mostly because the tabcontrol is not a direct ancestor, but the sibling of an ancestor)
Second we want to bind to the cursor position ( it needs to be updated after we add the text) and bind to the text... which doesn't work with this approach.
Third I tried to find a way to bind in xaml directly to a control of the the selected tab item without  much success...

Yannike: By the way, just realised that the start selection from a textbox is not bindable... lol

lzandman:
So what's the status on this? I see a branch conflict. Will this be solved. I'd like to have this functionality.

Yannike:
Haven't looked at it for a while, I could take another look

CodeMaid/Logic/Cleaning/FileHeaderLogic.cs
```diff
@@ -68,6 +73,68 @@ internal void UpdateFileHeader(TextDocument textDocument)
...
+
+                            break;
+                        }
+                    case "$PROJECT$":
+                        {
+                            try
+                            {
+                                if (_package.IDE.Solution != null)
+                                {
+                                    variableValue = _package.IDE.Solution.Projects.Item(1).Name;
```
codecadwallader: I believe this is just always going to give you the name of the first project in your solution.. not the project for the current file.  In other words if you have three projects in your solution A, B and C.. and your file is C.. shouldn't this resolve C?  The best way to approach that would be to walk upwards from the textDocument.ProjectItem

CodeMaid/UI/Dialogs/Options/Cleaning/CleaningUpdateDataTemplate.xaml
```diff
@@ -14,56 +16,63 @@
             </GroupBox>
             <GroupBox Header="{x:Static p:Resources.FileHeader}">
                 <StackPanel>
-                    <TabControl>
+                    <TabControl x:Name="LanguageTab" SelectedIndex="{Binding SelectedTabIndex}">
```
codecadwallader: The name `LanguageTab` can be removed.  Since we're using MVVM I try to avoid names so there isn't the temptation to directly manipulate view elements from the view model.

CodeMaid/UI/Dialogs/Options/Cleaning/CleaningUpdateViewModel.cs
```diff
@@ -52,6 +52,367 @@ public CleaningUpdateViewModel(CodeMaidPackage package, Settings activeSettings)
 
         #endregion Overrides of OptionsPageViewModel
 
+        #region Binding properties for variable header management
```
codecadwallader: Generally, this segment is massive (744 lines) and as far as I can tell it's all just to implement the ability to inject a header into the text box at the cursor location so the user doesn't have to type it or paste it from the clipboard?  It is a lot of code and all of the hard-coded indices are subject to break when/if we add more languages.  To be honest I think it would be better to just include the list of available variables as an informational message below the tabs and remove this code.  People have to type all of the content for the file header anyways, so it's actually probably slower to switch back and forth from the keyboard, to the mouse, and back to the keyboard.

CodeMaid/UI/SelectionBindingTextBox.cs
```diff
@@ -0,0 +1,105 @@
+﻿namespace SteveCadwallader.CodeMaid.UI
```
codecadwallader: Based on the comment above, I think this class can go away and the Options dialog can be reverted back to its previous behavior.

codecadwallader (changes requested):
Thank you again for your work on this feature.  I still have some concerns with the complexity of the UI and saw some issues with the replacement logic.  Please let me know if you have any questions!

CodeMaid/Logic/Cleaning/FileHeaderLogic.cs
```diff
@@ -58,6 +61,8 @@ internal void UpdateFileHeader(TextDocument textDocument)
                 return;
             }
 
+            settingsFileHeader = ReplaceVariables(textDocument, settingsFileHeader);
+
             var cursor = textDocument.StartPoint.CreateEditPoint();
             var existingFileHeader = cursor.GetText(settingsFileHeader.Length);
```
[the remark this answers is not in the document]
Yannike: Is it safe to assume that if a file starts with a comment it is probably a header and we don't change anything ?

CodeMaid/Logic/Cleaning/FileHeaderLogic.cs
```diff
@@ -68,6 +73,68 @@ internal void UpdateFileHeader(TextDocument textDocument)
...
+                // If the file header doesn't contain this variable, do nothing...
+                if (!settingsFileHeader.Contains(variable))
+                {
+                    continue;
+                }
+
+                var variableValue = string.Empty;
+                switch (variable)
+                {
+                    case "$USER_LOGIN$":
```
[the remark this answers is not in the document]
Yannike: Yep I'll remove it for now and will add it back when we have a framework checked in

CodeMaid/Logic/Cleaning/FileHeaderLogic.cs
Yannike: Yep you are right, I'll fix that! Thanks for spotting that !

CodeMaid/Logic/Cleaning/FileHeaderLogic.cs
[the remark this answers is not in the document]
codecadwallader: That's a good shortcut, and you could simplify it further to just look for the same few starting characters since some languages have multiple possible comment prefixes (e.g. `//` or `/*`).  However anytime someone renames a file or changes the header content, the header is going to out of date and won't be updated.

The old static approach simply inserted the new header anytime it was not already present as expected.  It had the draw back that if someone changed the file header, all files would get the new header and the old headers would have to be manually deleted (e.g. find/replace).  If we stuck with that approach then the new correct header would be inserted over the top of the old outdated header.  I don't think it's the greatest experience, but it's predictable and we definitely want to avoid any scenario of trying to auto-delete an old header (better to leave the code twice then risk losing it).

Thoughts on just going back to the previous logic of inserting whenever the calculated header isn't already present?

CodeMaid/Logic/Cleaning/FileHeaderLogic.cs
```diff
@@ -58,16 +61,72 @@ internal void UpdateFileHeader(TextDocument textDocument)
...
             {
                 cursor.Insert(settingsFileHeader);
                 cursor.Insert(Environment.NewLine);
             }
         }
 
+        /// <summary>
+        /// Replace the variables in the header if any
+        /// </summary>
+        /// <param name="textDocument"></param>
```
codecadwallader: Please finish filling out the XML documentation (e.g. settingsFileHeader should be included, textDocument should be defined).

CodeMaid/Logic/Cleaning/FileHeaderLogic.cs
```diff
@@ -58,16 +61,72 @@ internal void UpdateFileHeader(TextDocument textDocument)
...
+                // If the file header doesn't contain this variable, do nothing...
+                if (!settingsFileHeader.Contains(variable))
+                {
+                    continue;
+                }
+
+                var variableValue = string.Empty;
+                switch (variable)
+                {
+                    case "$SOLUTION$":
```
codecadwallader: Since these strings are repeated in a couple places, please make them constants.  The _variables array can be built from the constants.

CodeMaid/Logic/Cleaning/FileHeaderLogic.cs
```diff
@@ -58,16 +61,72 @@ internal void UpdateFileHeader(TextDocument textDocument)
...
+                        {
+                            if (_package.IDE.Solution != null)
+                            {
+                                variableValue = Path.GetFileNameWithoutExtension(_package.IDE.Solution.FullName);
+                            }
+
+                            break;
+                        }
+                    case "$PROJECT$":
+                        {
```
codecadwallader: These outer braces in a case statement are unnecessary.

CodeMaid/Logic/Cleaning/FileHeaderLogic.cs
```diff
@@ -58,16 +61,72 @@ internal void UpdateFileHeader(TextDocument textDocument)
...
+                            try
+                            {
+                                variableValue = textDocument.Parent?.ProjectItem?.ProjectItems?.ContainingProject?.Name ?? "";
```
codecadwallader: Please use `string.Empty` instead of `""`

CodeMaid/Logic/Cleaning/FileHeaderLogic.cs
```diff
@@ -58,16 +61,72 @@ internal void UpdateFileHeader(TextDocument textDocument)
...
+                            }
+                            catch (Exception)
+                            {
```
codecadwallader: Please add some diagnostic logging when the containing project cannot be found.  Example:
```
OutputWindowHelper.DiagnosticWriteLine("Unable to retrieve containing project", ex);
```

CodeMaid/UI/Dialogs/Options/Cleaning/CleaningUpdateDataTemplate.xaml
```diff
@@ -64,6 +66,13 @@
                             <TextBox Style="{DynamicResource TextBoxMultiLine}" Text="{Binding UpdateFileHeaderTypeScript, UpdateSourceTrigger=PropertyChanged}" />
                         </TabItem>
                     </TabControl>
+                    <GroupBox Header="Variables">
```
codecadwallader: Instead of a GroupBox, there's an existing UI convention of having an italicized text block underneath a TextBox.  Example:
```
<TextBlock FontStyle="Italic" Text="{x:Static p:Resources.PlaceEachRegularExpressionOnASeparateLine}" />
```
In this case I think $SOLUTION$, $PROJECT$ and $FILENAME$ would be independent of internationalization.  Also, why dollar signs instead of percents or other placeholders?

codecadwallader (changes requested):
Thank you for the updates, and no need to apologize - it took me awhile to get to this as well.  I think this is getting closer, and I think we can simplify it further.  Please let me know if you have any questions!

CodeMaid/Logic/Cleaning/FileHeaderLogic.cs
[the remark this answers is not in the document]
Yannike: Yep the problem I see with the old method is when we add other variables that varies like user or time. That won't work with the old static approach. I think if someone renames a file and he want the header to be updated, he just has to delete the header... not that big of a deal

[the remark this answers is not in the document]
Yannike: Yep the problem I see with the old method is when we add other variables that varies like username or date of creation. That won't work with the old static approach. I think if someone renames a file and he want the header to be updated, he just has to delete the header... not that big of a deal

Yannike:
@codecadwallader Here is the latest version, I think everything has been fixed ! :)

CodeMaid/Logic/Cleaning/FileHeaderLogic.cs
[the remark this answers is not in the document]
codecadwallader: Yeah, agree.  So will you please revert the logic below and go back to just checking for an exact match?  Then you can delete the two GetCodeLanguageComment methods as well (the less code the better).

CodeMaid.config
```diff
@@ -22,6 +22,10 @@
       <setting name="Feature_SettingCleanupOnSave" serializeAs="String">
         <value>True</value>
       </setting>
+      <setting name="Cleaning_UpdateFileHeaderCSharp" serializeAs="String">
```
codecadwallader: Lets take this setting override back out please.  I get the idea of not wanting to add file headers for C# and explicitly specifying it here so user preferences weren't inherited.. but that logic would extend to all 160+ settings so I'd prefer to keep it minimal.

codecadwallader: Although looking back over it.. there's a couple other old ones that should really be removed too.  Either way is fine, I'll clean up this file next time around.

codecadwallader (changes requested):
Looking really good, just one more request based on our conversation on updates!

CodeMaid/Logic/Cleaning/FileHeaderLogic.cs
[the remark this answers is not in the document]
Yannike: If you revert to the old method, things like name or date variable will always get replaced, I think the new method is better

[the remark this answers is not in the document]
codecadwallader: Ok, I'm willing to go with that.

I'm still not a fan of the hard-coded comment prefix list.  The file headers can be any kind of content (e.g. `using System;` or `#pragma`) so it would infinitely re-add if anything _except_ the expected comment prefix was present.  Additionally the `string.Empty` cases are always going to match since `str.StartsWith(string.Empty)` always returns true.

I think an alternative (and perhaps nice compromise) solution would be to still use the expected file header as the comparison point, but cut-off at the first dynamic element.  In other words before ReplaceVariables is called, create a method called GetStaticPrefix that would iterate across the variables, find the lowest IndexOf the variables and return the substring that matches up to that (defaulting to the full string when no variables are present).  Make sense?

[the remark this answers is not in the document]
Yannike: Yep, I'll add that, thanks

bregger:
@codecadwallader I am interested in getting the current year into the header for copyright purposes. I'd have it avoid overwriting the year in the header with the current year (doing so would mess up code clean up). Would that be a feature you'd support? We can discuss any design specifics you want to before I implement it.

IMO there isn't anything big here that doesn't need a bit of reworking, so I would implement my copyright year PR based on your feedback in this PR and the design I have in mind (more simple than what's here). 

Then if you want I could add the variables that are part of this PR in the same way, and we could close out this PR.

Honestly, I don't know why people want the file name, project name, or solution name in the file header. In fact, I would feel morally convicted to fight somebody to the death if they told me I needed it for something. But then again I felt the same way about the copyright year until I picked a fight with a lawyer so now I'm trying to automate adding it so our team can continue using codemaid.

codecadwallader:
@bregger lol, I appreciate your moral convictions. ;)  If you'd like to finish out this PR and add an additional variable in the form of a year I'm certainly open to that.  Would it auto-fill with the current year if it doesn't already exist?  I'm also curious to know a bit more about that lawyer fight.  We used to include a large copyright header at the top of every file, but I eventually dropped them as it seemed to fall out of public favor and I saw some unofficial references that it didn't have legal bearing.

bregger:
Yup, I'd make a variable that is filled in with the current year when the header is first added and then not updated on subsequent format runs. I'd probably call it CREATEDYEAR to make the intention clear. That would leave it open for MODIFIEDYEAR or the like to be added without harming backwards compatibility.

As far as the legal reasons for a copyright header - the explanation I was given was that the main purpose of the header is to make it clear who owns the code and what the license is. In our case, establishing in the source that the company is the sole copyright holder and that it is proprietary.

codecadwallader:
Forwards link to #667 which is an alternative approach to the same feature.

PR #519, opened 2018-01-23
Simplify vsct by removing unnecessary guids

heku:
Hi, before I pull a request for menu visibility, I'd like to pull this request for simplifying vsct.


codecadwallader:
Thank you for the contribution.  Can you explain why all those GUIDs were unnecessary and can all be identical?

heku:
Because VS identify an UI item (menu,button,group...) by both GUID and ID, so it's no need to give every individual UI item a GUID. For one package, it's better to define only one GUID for all UI item to avoid mess.

codecadwallader:
Ok that makes sense, thanks for teaching me that. :)  I had always assumed it needed to be unique per item but I checked a couple other popular repos and they also use a shared GUID.

codecadwallader:
Merged!  Thank you again for making this simplification.  The VSCT approach for defining integration with the IDE is very verbose and any simplifications are definitely helpful. :)

heku:
You're welcome, your codemaid is a great work, I like it very much and I learned so much from your code as well.

Issue #350, opened 2016-10-06
Exclude // TODOs from comment alignment


BalazsArva:
### Environment
- Visual Studio version: 2013/2015 any edition
- CodeMaid version: 10.1
- Code language: C#
### Description

The cleanup command wraps all comments so that they don't exceed the specified upper limit of line width. I like this feature, however, I'd like it even more to be able to exclude certain comments, for example "// TODO" lines. This would be helpful because Visual Studio's Task List window only shows the actual line that starts with // TODO.

So if I have something like this:

```
// TODO: Remove
// this later.
```

Then the Task list window will only show "TODO: Remove", whereas I'd like to see the whole text.
### Steps to recreate
1. Add a long // TODO comment
2. Perform a cleanup
3. The TODO is now wrapped into several lines
4. View Task List window, only the first line of the comment is shown.
### Current behavior

Long TODO's are wrapped into several lines.
### Expected behavior

We should be able to exclude lines/comments from autowrapping based on patterns.

codecadwallader:
Thanks for the suggestion to make an exception case for TODO comments wrapping when comment formatting is invoked upon them.  You mentioned being able to setup patterns for exclusions, were you thinking regular expression based patterns?

BalazsArva:
Yes, though I am not sure if that wouldn't be too expensive. Maybe StartsWith() would do as well.

DanielMKelly:
Possibly the Codemaid/Helpers/CodeCommentHelper.cs function GetCommentRegex should exclude TODO from registering as a comment and therefore be exempt from all formatting?

willemduncan:
Agreed, same as with the `// Resharper Disable/Enable` comments. I would hard code a list and not expose it to configuration just yet.

willemduncan:
> By default, Visual Studio includes the following tokens: `HACK`, `TODO`, `UNDONE`, `NOTE`. These are not case sensitive.

Let's add all of them while we're at it. There is no requirement for a semicolon `:`, either in documentation or in my tests on VS2017.

Which brings the problem, how to distinguish between this which should be left alone:
```cs
// Some piece of comment.
// TODO Do something very important with a long description that should not be wrapped.
// Maybe another comment.
```
And this which should be formatted as usual
```cs
// Writing some text using the
// todo word in the middle of it
// without it being something special.
```

[...]

PR #545, opened 2018-04-26
Ignore comment lines starting with certain prefixes

### LINKED ISSUE #347: Format Comments breaks ReSharper code inspection configuration comments
### Environment
- Visual Studio version: Visual Studio 2015 Professional 14.0.25431.01 Update 3
- CodeMaid version: 10.1.93
- ReSharper version: ReSharper Ultimate 2016.1.2
- Code language: C#
### Description

CodeMaid's 'Format Comments' command concatenates/splits comments used by ReSharper for code inspection configuration, causing them to not be recognized by ReSharper.
### Steps to recreate
1. Add ReSharper code inspection configuration comments in a C# source code file.
   // ReSharper disable UnusedMember.Global
   // ReSharper disable MemberCanBePrivate.Global
   // ReSharper disable UnusedAutoPropertyAccessor.Global
   // ReSharper disable AutoPropertyCanBeMadeGetOnly.Local
2. Run the 'format comment' command on the block comments.
### Current behavior

ReShaper allows the use of comments to enable/disable/configure code inspections in the source code file.  There are inserted with one option per line.  When 'Format Comment' is run on a block of these comments, it concatenates them together based on the 'Wrap comments at column' value.  Doing so causes ReSharper to not recognize them.  _This probably only effects folks who enable the 'Run format comments during cleanup' option, as it is unlikely that anyone would explicitly run the 'Format Comment' command the ReSharper block._
### Expected behavior

Don't break things / do not format comment lines which contain ReSharper code inspection configuration options.

My CheapHack(tm) fix would be to have CodeMaid.Helpers.CodeCommentHelper.IsCommentLine return false if the line starts with "// ReSharper disable" or "// ReSharper enable".

willemduncan:
Fixes #347, #350.

Ignore (ie output literally) any comment line starting with certain prefixes while formatting. Updated the comment regex, matching, and formatter. To prevent adding even more arguments to the function calls, two options classes have been added which are now used as arguments instead.

Lines starting with any of the tokens defined in the options for the Task List Window will be ignored. The tokens must be followed by a semicolon and a space to qualify, eg `TODO: Lorem ipsum`. Matching is case sensitive.

In addition, there is a new entry in the settings file to defining other tokens to ignore. For the moment, this only contains `ReSharper disable` and  `ReSharper enable`. Once again, this is case sensitive.

Added a few unit tests to cover some obvious cases.

### DIFF
--- a/CodeMaid.UnitTests/CodeMaid.UnitTests.csproj
+++ b/CodeMaid.UnitTests/CodeMaid.UnitTests.csproj
@@ -112,6 +112,7 @@
     <Compile Include="Formatting\CommentFormatHelper.cs" />
     <Compile Include="Formatting\HeaderFormattingTests.cs" />
     <Compile Include="Formatting\FormatWithPrefixTests.cs" />
+    <Compile Include="Formatting\IgnorePrefixesTests.cs" />
     <Compile Include="Formatting\SimpleFormattingTests.cs" />
     <Compile Include="Formatting\XmlFormattingTests.cs" />
     <Compile Include="Formatting\ListFormattingTests.cs" />
--- a/CodeMaid.UnitTests/Formatting/CommentFormatHelper.cs
+++ b/CodeMaid.UnitTests/Formatting/CommentFormatHelper.cs
@@ -38,11 +38,7 @@ namespace SteveCadwallader.CodeMaid.UnitTests.Formatting
 
         public static string Format(string text, string prefix)
         {
-            var xml = XElement.Parse($"<doc>{text}</doc>");
-            var line = new CommentLineXml(xml);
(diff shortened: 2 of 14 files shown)


codecadwallader:
Thanks for putting this together, it looks good to me and will merge in.

In general I think anything that is a setting has been exposed through the Options dialog, so it may be useful to add that for the IgnoreTokens as well for consistency.

willemduncan:
I prefer to be on the "Joel Spolsky side" of things: don't bother users with too many settings ;) Yet I didn't want to hardcode ReSharper specific exceptions. As soon as more exceptions show up, or users request the ability to ignore more prefixes, we can either expose the settings or explain them how to manually edit the config. 

Thanks for merging!

PR #547, opened 2018-04-30
Enable compilation with the latest version of the C# language.

codecadwallader:


willemduncan:
Is it a conscious choice to use `latest` over the explicit current `<LangVersion>7.1</LangVersion>`? 

For example, version 7.1 has a breaking change over 7.0 because of the "implicit tuple name" feature. If one developer has 7.0 installed, and the other 7.1, this will cause issues that an explicit version will prevent.

codecadwallader:
That's a fair point and thanks for raising it.  Yes it was a deliberate choice, primarily so that we don't have to make updates every time a new minor version is released.  I believe the default behavior when unspecified is implicitly the latest major version which would have the same potential to be different on different developer's machines.  I'm inclined to let it go as-is but would certainly adjust if it became a common problem.

PR #550, opened 2018-05-03
Sort lines in same order so that we can compare them

heku:
It's better to sort all .resx files in same order, so that we can compare them like below.
capture


maikebing:
It is recommended to use the[ ResX Resource Manager ](https://marketplace.visualstudio.com/items?itemName=TomEnglert.ResXManager)if only for translation convenience

codecadwallader:
I agree about sorting the .resx files for comparison purposes.  I have done the same with the App.config as an example to make it easier to diff across versions.  Thank you!

heku:
Yes, exactly. However, it would be easier for comparing them if we sort them in same order.

Issue #263, opened 2016-04-08
CodeMAID Spade: Options to Display Type Information


kgallaha:
**Split from #259**

This is a feature request to add settings to display type information next to all class members.  This would include fields, properties, and methods.  It could be presented pretty simply as follows:

$MemberName$ : $TypeName$

If simple to implement, one could also CTRL+Click the type name to navigate to it's definition.

codecadwallader:
Thank you for the suggestion.

PR #582, opened 2018-08-17
Spade: Option to Display Type Information, #263, disabled by default

GammaWolf:


codecadwallader:
Thanks for the PR.  Overall it looks good.  :)  Two suggestions:

1.  Instead of calling it "Show Types" how about "Show Return Types" since it is just for controlling the visualization of return types?  We already show types for arguments always.
2.  We use a different display style for argument types.. should we considering using the same for return types?

GammaWolf:
I have made the changes.

1. I had also noticed the naming issue but wasn't sure to include 'return' since it also controls the visualization of fields and events which I wouldn't see as having return types. But it's probably better having it than having the conflict with it not controlling the types in the arguments.

codecadwallader:
1. Yeah that's a good point.  The underyling EnvDTE generally just calls it the "type" of the item.  Looking at the other options perhaps "Show item type" would be more accurate?

I noticed that in the converter we're directly reading from settings, but before the converter was settings agnostic and instead had properties/overloaded default instances for the different settings choices.  I think either way works, but I'd like to have it be the same so there's not two different ways one converter is getting settings state.  I can make these changes if that isn't clear.

codecadwallader:
I've simplified the XAML and have the converter reading from settings.  I also changed the setting name to "Show item types" and enabled it by default.  Do you want to take a look back over it and let me know if you see any other changes needed before we merge?

GammaWolf:
Looks good to me.

PR #593, opened 2018-10-23
Switcher: Added .cshtml -> .cshtml.cs support

derekmckinnon:
Added another entry in the `Settings.settings` file that allows easy switching between Razor Page views and PageModels (if present).


codecadwallader:
Looks good - thanks for adding that! :)

PR #617, opened 2019-01-27
New splashscreen integration

kdrapel:
New splashscreen integration (Paint.NET file with layers available, font is Bahnschrit). 10.7 set as the version.


kdrapel:
image

codecadwallader:
Thank you for creating this image, it looks very nice.  One concern I have is it doesn't match our existing design (e.g. codemaid.net)

kdrapel:
Thanks for your feedback, will rethink about it. Did not have the website in mind when doing it. Will try to design a version with a gray background and greenish tones

xmedeko:
If you hard code version and copyright year into the PNG, you have change PNG every time you bump the version of change the copyright year. Better to draw these in XAML - and that affects the font used, too.

codecadwallader:
@xmedeko Yep you are correct.

kdrapel:
> If you hard code version and copyright year into the PNG, you have change PNG every time you bump the version of change the copyright year. Better to draw these in XAML - and that affects the font used, too.

The PDN file is available, you can remove the fonts, export a PNG and use XAML.

PR #613, opened 2019-01-06
Remove blank newline in self-closing doc comment

bonimy:
## Issue

Ending a doc comment with a self-closing XML tag will cause CodeMaid to add a trailing comment line.

### Example

CodeMaid will format a document to look like the following
```csharp
/// <seealso cref="IFormattable.ToString(String, IFormatProvider)"/>
///
public static string GetUIString(IFormattable value, string format = null)
{
    return value?.ToString(format, CurrentUICulture);
}
```

## Resolution

Check whether a self-closing tag was the end of the XML comment block, and skip adding a new line if it is. Fixes #599.

### Example

CodeMaid will format a document to look like the following
```csharp
/// <seealso cref="IFormattable.ToString(String, IFormatProvider)"/>
public static string GetUIString(IFormattable value, string format = null)
{
    return value?.ToString(format, CurrentUICulture);
}
```

## Tests

I could not set up unit tests in my environment, so I have no guarantees that this works outside of the test runs I did on my own projects.

Issue #599, opened 2018-11-14
Adds comment marks after self-closing xml documentation elements.


Mythilt:
### Environment
- Visual Studio version: 2017 - 15.8.9
- CodeMaid version: 10.5
- Code language: visual basic

### Description
  Comment marks are added to the line after any self-closing xml documentation tags during cleanup, causing potential documentation errors.
  Apparently a similar issue was fixed with regards to C# /// xml-documentation comments in version 0.9.1 or so (issue #215).
  This may be a side-effect of setting the 'Always split all xml tags onto multiple lines' formatting setting.  Considering that self-closing tags have no &lt;/ tagname&gt; to move to the new line, that may be the cause of this problem.

Example 1:
Before cleanup.
''' &lt;summary&gt;
'''   Class SomeService
''' &lt;/summary&gt;
''' &lt;seealso cref="System.ServiceProcess.ServiceBase" /&gt;

After cleanup
''' &lt;summary&gt;
'''   Class SomeService
''' &lt;/summary&gt;
''' &lt;seealso cref="System.ServiceProcess.ServiceBase" /&gt;
'''  &lt;&lt;&lt; Issue is here.

Example 2:
Before cleanup.
''' &lt;summary&gt;
'''   Class SomeService
''' &lt;/summary&gt;
''' &lt;seealso cref="System.ServiceProcess.ServiceBase" /&gt;
''' &lt;remarks&gt;
'''   This is an example.
''' &lt;/remarks&gt;

After cleanup
''' &lt;summary&gt;
'''   Class SomeService
''' &lt;/summary&gt;
''' &lt;seealso cref="System.ServiceProcess.ServiceBase" /&gt;
'''''' &lt;remarks&gt;      &lt;&lt;&lt; Issue is here.
'''   This is an example.
''' &lt;/remarks&gt;

### Steps to recreate
1. Create documentation xml comments with a self closing tag (e.g. &lt;customtag /&gt;
2. Run Codemaid format command.
3. Self - closing document will have ''' added to line after the tag, potentially causing a formatting issue.

### Current behavior
  By adding a new comment line after the self-closing xml tag, any following comment lines may have a formatting issue (if more than three ' characters, visual studio assumes the xml documentation comment is itself, commented out.)

### Expected behavior
  Self-closing xml tags should not add a new comment line after the tag.

codecadwallader:
Thanks for reporting the issue.  @willemduncan can you please take a look when you have a chance?

[...]

Issue #564, opened 2018-06-11
CodeMaid 10.5 broke whitespace handling


alexdrl:
### Environment
- Visual Studio version: VS 15.7
- CodeMaid version: 10.5
- Code language: C#

### Description
PR #546 has broken whitespace handling in some scenarios. This test reproduces this result:
```cs
        /// <summary>
        /// If XML tag indenting is set, this should not affect any literal content. however,
        /// content after the literal should be indented as normal.
        /// </summary>
        [TestMethod]
        [TestCategory("Formatting UnitTests")]
        public void XmlFormattingTests_DoesIndentAfterLiteralContent2()
        {
            var input =
               "<summary> Return the model in XML format. </summary>" + Environment.NewLine +
               "<remarks>" + Environment.NewLine +
               "This method is cloned from the <see cref=\"DomainDSLSerializationHelperBase\"/> generated class." + Environment.NewLine +
               "The purpose of this method is to add the" + Environment.NewLine +
               "</remarks>";

            var expected =
               "<summary> Return the model in XML format. </summary>" + Environment.NewLine +
               "<remarks>" + Environment.NewLine +
               "    This method is cloned from the <see cref=\"DomainDSLSerializationHelperBase\"/> generated class." + Environment.NewLine +
               "    The purpose of this method is to add the" + Environment.NewLine +
               "</remarks>";

            Settings.Default.Formatting_CommentXmlValueIndent = 4;
            Settings.Default.Formatting_CommentXmlKeepTagsTogether = true;
            Settings.Default.Formatting_CommentXmlSpaceTags = true;

            // First pass.
            var result = CommentFormatHelper.AssertEqualAfterFormat(input, expected);

            // Second pass.
            CommentFormatHelper.AssertEqualAfterFormat(result, expected);
        }
```
A line get created after the remarks line. also, there are a lot of tests also failing...

codecadwallader:
Thanks for reporting the issue.  In regards to failing tests, there's a known issue #467 with the integration test suite which is the bulk of the tests.

@willemduncan want to take a look at the whitespace issue mentioned?

alexdrl:
@codecadwallader The tests that failed in my environment were the UnitTests Project, which I thought that they will be working. For executing them, I deselected the runsettings file, so they won't got ignored.

[...]

Issue #413, opened 2017-03-17
Feature Request: User defined template for comment formatting


bspraguetx:
### Environment
- Visual Studio version: VS 2015
- CodeMaid version: 10.2
- Code language: C#

### Description
The comment formatting that is currently being done meets our needs for just about everything but we do have a couple of standards that are not supported out of the box.  For example, we ask that our developers split the `<return>` and `<remarks>` tags into multiple lines similar to option provided for the `<summary>` tag.  We also keep a change log in the code files that consists of a `<history>` tag with multiple child `<change>` tags.  My thought was that this could be addressed if you could provide the ability to define a template for additional formatting.  Something like:

```
<return>
[[Content]]
</return>
<history>
[[Content]]
</history>
    <change>
    [[Content]]
   </change>
```

codecadwallader:
Thanks for the suggestion.  The closest existing setting we have is "Always split all XML tags onto multiple lines" which could be a good starting point for anyone interested in pursuing this request.

surekqomi:
I have a similar problem but with the `<summary>` tag. Our company has a new coding standard that indents all tags _except_ the `<summary>` tag, which means manually removing the indent added by the `Extra indent for values inside XML comments` option. Since we already had a sizable code base before this new standard was introduced, there is a very large amount of reworking taking place, and very often this non-indent on `<summary>` is missed causing inconsistency.

willemduncan:
This would come down to allowing a user definable indent and splitting for each XML tag individually, including the option to add custom tags (as far as I know `<history>` and `<change>` are non-standard). Would that meet the requirements?

bspraguetx:
Agreed

Brad

PR #620, opened 2019-02-11
Individual tag formatting rules

### LINKED ISSUE #364: Feature request - Setting wrapping comment column on-the-fly
Hi there,

First, thanks for a great product. I'm the author of TranSolution (http://www.hexadigm.com) and really appreciate your efforts. Not sure if this is where feature requests go but I've got one that *may* be fairly simple to implement. I'm forever wrapping comments via <Ctrl+M, N> and would like to be able to quickly change the wrapping column on-the-fly, without having to launch settings all the time. It would be nice if you could add, say, a "Prompt each time" checkbox under the "Wrap comments at column" setting. It could be off by default for backwards compatibility but when checked, <Ctrl+M, N> would then display a small dialog with a single field for entering the column on-the-fly (initialized to the last value entered on this dialog which would initially be the "Wrap comments at column" setting itself). This would then override the "Wrap comments at column" setting so once OK is pressed (on this new dialog), wrapping would then use the newly entered column value instead of the one from settings (and the new value will also be used to initialize the same dialog next time I press <Ctrl+M, N>). You could even add (time permitting) a few extra check boxes on this dialog if you want, such as an "Update settings with this new value" check box (if checked then the "Wrap comments at column" setting itself will be updated), a "One time only" check box (if clear, the newly entered value will be used to initialize the same dialog next time but if checked, then it won't), etc. You'ld have to think about some of this of course (I haven't thought through all the details), but it would be a very useful feature (and maybe fairly quick to implement but only you know that of course). As an alternative, if you want to get really fancy (as if you have the time :), you can even consider wrapping all text using the right-most column of the currently selected text block. That is, if I highlight a block of text (normally comments), as currently described here https://msdn.microsoft.com/en-us/library/729s2dhh(v=vs.100).aspx (Google for "How to: Select and Change Text Visual Studio" if the link is now dead - quotes not included of course), in particular a rectangular block of text using "Column (Box) Mode" (search for this in the latter link), it would be nice if I could then format everything in that block using some hot key (or whatever). No column number would then have to be entered since the block itself defines the wrapping area (the right-most column anyway). In any case, some way to quickly set the wrapping column on-the-fly would be very useful. On the surface the first idea above in particular (or some hybrid) hopefully wouldn't take much effort. Thanks very much for your consideration (appreciated).

willemduncan:
This has been hiding on my laptop for quite a while now. With all the whitespace and newline problems of the last year, the comment formatting code was becoming more and more unmaintainable. This PR addresses a lot of issues and a few design problems. 

The biggest improvement is that there is no more magic strings and exceptions hiding in the code. It's all moved to the settings. 

- The settings have been restructured and get passed into the formatter. This is the base for being able to fix #364. 
- Settings now include a dictionary of XML tags, where each tag can have its own rules overriding the default. This is the base for fixing #413.
- I have re-enabled the unit test related to #564, it passes now.
- It includes the fix for #599, therefore probably conflicts with PR #613. Sorry @bonimy, I took your solution and applied it to the newly structured code.
- It adds a few new unit tests to ensure compatibility.

There is one big elephant in the room though: I do not know enough XAML to come up with a GUI supporting all the new possibilities. Right now, there is a little wrapping code that translates the old (current) style settings into the new settings object. It works, but hides all the nice new features. I do feel however, that this code is a big step forward even without updated GUI.

It's a shame to let this code go old and see other PRs addressing issues that are already fixed here.

This will need some rigorous testing though. The unit tests all pass, but do not properly cover all the cases, as proven in #599.


willemduncan:
Interesting, the AppVeyor build has one unittest fail:
```
Failed   XmlFormattingTests_AddSpaceToTagContentShouldLeaveNoTrailingWhitespace2
Error Message:
 Assert.AreEqual failed. Expected:<<remarks>
    Lorem ipsum dolor sit amet, consectetur
    adipiscing elit.
</remarks>>. Actual:<<remarks> 
    Lorem ipsum dolor sit amet, consectetur
    adipiscing elit.
</remarks>>.
```
But I do not see why, and it does not fail on my machine.

codecadwallader:
Whoops, I already merged #613 in since I thought it had your blessing.  I suspect that caused the conflict.  Do you need me to roll it back, or can we use that fix in the interim while working on this larger change?

How can we proceed with the rigorous testing you mentioned?

On a different note, VS2019 is scheduled to officially release on April 2nd so I'm hoping to do a significant change to switch us to the AsyncPackage model (see #563) which will break backwards compatibility (VS2017+ only) and be the next major version (v11).  I'll be pretty heads down trying to make that happen.

willemduncan:
No problem, I will fix the conflict. Leave the fix in, credit where credit is due.

Is there still an option to run integration tests, or are they still not working?

I'm working on a follow up for this PR, where I finally separate the responsibilities of the various formatting classes. This improves testability for the future. Somehow even with our current set of tests, issues seem to slip by sometimes.

VS2019 support (and async) is more important,

codecadwallader:
Sounds good, thanks.  And yeah unfortunately integration tests are still not functioning (#467)

germandb:
Hi @codecadwallader  VS 2019 was already released any idea when this PR will be integrated.
I'm in need of the FIX for the #599 and #413.

codecadwallader:
@germandb I'll have to punt to @willemduncan for comment formatting functionality questions.

willemduncan:
I'm here. This can be merged, it works on my machine™️ and as said the tests pass. I would suggest a RC/Beta build to go out though before pushing it into the stable build.

codecadwallader:
Thanks Willem. :)  Originally you mentioned:

> There is one big elephant in the room though: I do not know enough XAML to come up with a GUI supporting all the new possibilities. Right now, there is a little wrapping code that translates the old (current) style settings into the new settings object. It works, but hides all the nice new features. I do feel however, that this code is a big step forward even without updated GUI.

Do you still feel the same way about moving forwards with the changes as-is?

willemduncan:
Yeah, I know the elephant is still there. 

I do not see myself creating a new GUI in the near future, plus we haven't really figured out how to store more complex configuration options, things like dictionaries. Especially since it should have really have compatibility with `.editorconfig`.

For now there is a mapping from the existing settings to the new `options` object ([see code here](https://github.com/codecadwallader/codemaid/pull/620/files#diff-7c252dbea61d31044d7b8e5ef85d3cefR63)), which is handled on the fly. It's suboptimal but it is only a single location that needs to be changed in the future. It's the best compromise I can think of for now.

So yes, I would go ahead and merge. I'll be there the coming weeks to quickly catch any fallout, should it occur.

codecadwallader:
Alright cool, I'll merge it in now and it will be available in the CI channel shortly if you or @germandb want to check it out!

germandb:
Ok, I will try the new release this week and  give you feedback.

PR #647, opened 2019-04-18
Fix magically added slashes

### LINKED ISSUE #646: Comment formatting incrementally adds slashes
We've tested the v11.0.188 build from the CI. Both issues seem to have resolved, but a new one seems to have arised... 😅

When formatting a comment with at least a section that is multiline, it starts adding backlashes each time is formatted.

For example:

Original comment
```cs
/// <summary> This is a summary. </summary>
/// <param name="param"> This is a multiline comment for the parameter param. It is going to add backlashes. </param>
/// <returns> The returns. </returns>
```

Then, after the first time I format the document:
```cs
/// <summary> This is a summary. </summary>
/// <param name="param">
///     This is a multiline comment for the parameter param. It is going to add backlashes.
/// </param>
///        /// <returns> The returns. </returns>
```

A second time:
```cs
/// <summary> This is a summary. </summary>
/// <param name="param">
///     This is a multiline comment for the parameter param. It is going to add backlashes.
/// </param>
///        /// ///
/// <returns> The returns. </returns>
```

From then on, the line of backlashes will grow each time the document is formated.

Should we keep using this issue or create a new one?

_Originally posted by @Starkie in https://github.com/codecadwallader/codemaid/issues/564#issuecomment-483533240_

willemduncan:
Once again, I cannot seem to fix an issue without reorganizing half the formatting code. 

Up to now, a call to `NewLine()` created a newline and immediately wrote a comment prefix (eg `///`). Then indenting was added manually from the individual line formatter. 
I've changed this to add the prefix and optional indenting when the first word is added to the output, which makes for a more logical and less error prone process. It's still messy, but slightly less so 😉 

At least this fixes #646.

### DIFF
--- a/CodeMaid/.editorconfig
+++ b/CodeMaid/.editorconfig
@@ -1,0 +1,8 @@
+; Top-most EditorConfig file
+root = true
+
+[*.{cs,vb}]
+dotnet_style_qualification_for_field = false:suggestion
+dotnet_style_qualification_for_property = false:suggestion
+dotnet_style_qualification_for_method = false:suggestion
+dotnet_style_qualification_for_event = false:suggestion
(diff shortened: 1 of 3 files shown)


CodeMaid/.editorconfig
```diff
@@ -0,0 +1,8 @@
+; Top-most EditorConfig file
+root = true
```
codecadwallader: Cool!

PR #670, opened 2019-08-12
Ensure imported Config File is not read-only

Smartis2812:
Had a little bug, where the config file was copied from a global storage and setted read-only.
First time it works, but if i tryed to re-import the file i got an exception because the target file is now read-only.
This fix should ensure that the imported file is always not read-only and keeps all other file attributes from the original config.

### DIFF
--- a/CodeMaid/UI/Dialogs/Options/OptionsViewModel.cs
+++ b/CodeMaid/UI/Dialogs/Options/OptionsViewModel.cs
@@ -283,6 +283,7 @@ namespace SteveCadwallader.CodeMaid.UI.Dialogs.Options
                 try
                 {
                     File.Copy(dialog.FileName, ActiveSettingsPath, true);
+                    File.SetAttributes(ActiveSettingsPath, File.GetAttributes(ActiveSettingsPath) & ~FileAttributes.ReadOnly);
 
                     RefreshPackageSettings();
 


codecadwallader:
Thanks for the PR! :)  It makes sense that the attributes of an imported file are being copied across (vs. the content which is the true intent).  Are there any other file attributes that you think could be a problem on import?

Smartis2812:
I have written a short example for testing if you can delete a File with all .net available FileAttributes.
So far only a read-only File causes an exception. After this test I would say the other FileAttributes shouldn't cause any trouble, so far. 

```
static void Main(string[] args)
{
    string exampleDataFile = @"ExampleData.xml";
    string targetPath = @"C:\Temp\";

    foreach (FileAttributes fileAtt in Enum.GetValues(typeof(FileAttributes)))
    {
        try
        {
            string targeFilename = string.Format("{0}ExampleData_{1}.xml", targetPath, fileAtt);
            File.Copy(exampleDataFile, targeFilename);
            File.SetAttributes(targeFilename, fileAtt);
            File.Delete(targeFilename);
        }
        catch (Exception ex)
        {
            Console.WriteLine("{0}: {1}", fileAtt, ex);
        }
    }
}
```

codecadwallader:
Cool, thanks for running that test.  Merging in now and thanks again for this improvement! :)

PR #671, opened 2019-08-20
Added persistence for `IsExpanded` state of the `CodeItemRegion` node on switching active document…

oleg-shilo:
It is a user-experience improvement CR.

_*Problem*_
`IsExpanded` state of the nodes is lost after switching the active document. 

While seems like a very minor problem, it does lead to the navigation difficulties with the C# code containing moderate-to-large number of insignificant (from business logic point of view)  members. 

For example a simple _view model_ class with plenty of full bodied properties (and nothing else) will have 50% of the Spade real estate occupied (wasted) by the fields, which have no real business value in this case. For a typical "data record" class only properties are meaningful.

The problem is easily handled by collapsing the _Fields_ node. But currently after switching the active document the _Fields_ node gets expanded again.

_*Solution*_

Allow maintaining the node `IsExpanded` state when switching the active document.


codecadwallader:
Thanks for putting this together.  It's an interesting idea and I have a couple questions:

1.  It might surprise some users when something they collapsed in one file is automatically collapsed in another file.  Would it make more sense if the state was remembered per file?
2.  It sounds like you might always want certain regions to be collapsed.  Would it make sense to have options to always collapse regions with certain names or perhaps based on size?

oleg-shilo:
Hi Steve,

The answer to both of your questions "yes it would"  :)

With this very PR only wanted to present you the solution approach. And... if it is accepted in principle I was planing to esquire about the appropriate persistence model (per file or/end global).

> ... Would it make more sense if the state was remembered per file?
Please let me know if you prefer I can extend the solution to remember the `IsExtended` per file and I will update the PR.  

> Would it make sense to have options to always collapse regions with certain names or perhaps based on size?

I actually though about a bit simple cofig option. Why don't CodeMaid has a global initial `IsExtended` state of the nodes (fields, props, methods...), which is applied on opening the file for the first time.  And after that it is maintained as per file? What do you think?

codecadwallader:
I like the idea of having an option where users can specify regions to collapse by default.  I think it could be inside the "Collapsing" area inside CodeMaid->Options and have a similar text box approach to the "Switching" area.  I think this would be a good start and probably cover your use case by itself since you could specify that regions with the name "Fields" (or whatever it is called) should always be collapsed by default.

oleg-shilo:
Great. Then we are on the same page.

Then the next question. :)

Is it something that is better to be implemented by yourself since having this in options would affect the app-config user experience. And arguably you know this area of code better. Or it is something that you prefer to be contributed.

codecadwallader:
It is open for contribution.  Hopefully the existing example of using a list of options in switching (and there's another in the same style under cleaning exclusions) are helpful. :)

oleg-shilo:
Great, I will try to find the time to rework the PR then.

One question though. Where can I find that "existing example of using a list of options in switching"?

codecadwallader:
Sure, the "Switching" side panel in Options has a similar option that allows a user to type in multiple values delimited by new lines.  So for example one line has ".cpp .h" and a second line may have ".xaml .xaml.cs".  I think this same UI pattern could be used to allow users to specify "Fields" "Variables" or any other region names they want to treat specially.

UI: https://github.com/codecadwallader/codemaid/blob/master/CodeMaid/UI/Dialogs/Options/Switching/SwitchingDataTemplate.xaml

ViewModel:
https://github.com/codecadwallader/codemaid/blob/master/CodeMaid/UI/Dialogs/Options/Switching/SwitchingViewModel.cs

Logic:
https://github.com/codecadwallader/codemaid/blob/master/CodeMaid/Integration/Commands/SwitchFileCommand.cs

oleg-shilo:
It took all 3 month for me to get back to this PR. And after re-reading your comments I have to conclude that we lost it at the second post in this thread :)

Now I am convinced that we were discussing two completely different things. The PR was about **Spade** view remembering `IsExpended` state of the individual nodes. 
    _'...full bodied properties (and nothing else) will have 50% of the Spade real estate occupied (wasted) by the fields...'_

And it looks like you got the impression that the change is about collapsing the code regions in the IDE document:
     _'I think it could be inside the "Collapsing" area inside CodeMaid->Options...'_

The fact that I mentioned "Spade" only once did not help the confusion. And then we started discussing the best place for reflecting the feature in user settings and... completely lost the fact that we were not yet talking about the same thing. :) 

Anyway, to restart the discussion I have created a new PR where properly described the feature and its objective. I have also added another feature "Hiding Spade nodes" that addresses the same problem of Spade overcrowding with the large files.

-------

Though I do not want to dismiss your consern just in case if it was not caused by misunderstanding and was rather about user experience with Spade content.


    > @codecadwallader: It might surprise some users when something they collapsed in one file is automatically collapsed in another file. Would it make more sense if the state was remembered per file? 
    
    
Actually it should not surprise any user as it is the user experience Visual Studio is promoting in all "tree-view" content presenters. Thus file properties view is being updated everytime user opens a new file and yet the `IsExpanded` state of the all nodes is maintained globally through the while IDE user session:

image

Issue #551, opened 2018-05-04
remove and sort namespaces in .XAML files


Apflkuacha:
### Environment
- Visual Studio version: 2017 Community
- CodeMaid version: 10.4
- Code language: C#

### Description
The function "Remove and sort namespaces" of VisualStudio is not executed at .XAML files when using cleanup active document. This should be done or at least a setting for this would be great.

codecadwallader:
Thanks for the suggestion!

FletcherJohnsonZ:
I am having the same problem.  Any idea if this is or will be addressed?

codecadwallader:
It looks like XAML files have their own IDE command that would need to be invoked.  I believe it is `EditorContextMenus.CodeWindow.RemoveAndSort`.  Some regression testing and language distinction would need added.  Supporting re-insertion of specific namespaces would need some additional consideration. 
 https://github.com/codecadwallader/codemaid/blob/master/CodeMaid/Logic/Cleaning/UsingStatementCleanupLogic.cs is probably the main focus point of where that logic would go.

Apflkuacha:
I was motivated to take a look into that and made my first Pull request :) The UsingStatementCleanupLogic was never executed on XAML files, that's why this filetypes weren't considered. I hope this changes make it into the next release.

PR #692, opened 2019-12-02
Remove and sort namespaces in .XAML files. Fixes `#551`

Apflkuacha:
Call the RemoveAndSortUsingStatements() also on XAML files. I've tested the changes also with the other Markup filetypes (XML and HTML) and also with the CodeMaid setting "Cleaning - Visual Studio" "Using statements to re-insert when removed" and no problems occurred.


codecadwallader:
Awesome, thank you for doing the investigation and submitting this PR!  I'm merging it now and it will be in the next release.  You'll be able to check it out / use it in the interim if you utilize the CI builds that are here: http://vsixgallery.com/extension/4c82e17d-927e-42d2-8460-b473ac7df316/

PR #693, opened 2019-12-03
Add static sort activation/deactivation feature to reorganize

shuelsmeier:
Hi,

I added the possibility to include/exclude the ability to sort static classes, methods, etc.

- Include static sort active/inactive in settings
- Extend options/reorganize/types dialog to support changing the static sort behaviour

### DIFF
--- a/CodeMaid.UnitTests/MemberTypeSettingTests.cs
+++ b/CodeMaid.UnitTests/MemberTypeSettingTests.cs
@@ -9,7 +9,7 @@ namespace SteveCadwallader.CodeMaid.UnitTests
         [TestMethod]
         public void CanSerializeMemberTypeSetting()
         {
-            var memberTypeSetting = new MemberTypeSetting("Fields", "Member Variables", 1);
+            var memberTypeSetting = new MemberTypeSetting("Fields", "Member Variables", 1, true);
             Assert.IsNotNull(memberTypeSetting);
 
             var serializedString = (string)memberTypeSetting;
@@ -19,7 +19,7 @@ namespace SteveCadwallader.CodeMaid.UnitTests
         [TestMethod]
         public void CanDeserializeMemberTypeSetting()
         {
-            const string serializedString = @"Fields||1||Member Variables";
+            const string serializedString = @"Fields||1||true||Member Variables";
 
             var memberTypeSetting = (MemberTypeSetting)serializedString;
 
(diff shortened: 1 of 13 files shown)


codecadwallader:
Thanks for putting this together. :)  It'll take some time for me to go through it since it's larger.  Some initial questions that come to mind:

1.  There's an existing setting to push explicit interface implementation members to the end.  It doesn't allow per type setting levels though.  What do you think about this "all" approach vs. per-type approach?
2.  Do you have an example of when different types have different static sorting rules?
3.  Could/should(tm) the ability to "ignore static" be expanded to other attributes (e.g. type, access, explicit interface, constant and read-only)?

shuelsmeier:
In our codebase, we sort the methods of a class by access from public to private and then by their name. Due to performance issues private methods of the class are declared static whenever possible. When static and non-static methods are sorted by the current algorithm, declaring a method static obviously moves the method in the class to another position. This makes differences between two changesets different to spot.

I think that it is better to give the user the possibility to decide per-type. It can be that a user would like to have a static sorting for some types, but for others not. However I don't have a real world example.

The ability to "ignore static" could be expanded to other the attributes within a couple of hours. I would be willing to do that.

codecadwallader:
Thanks for the information.  :)

Since there isn't currently a need/request for per-type static rules, I think it puts a lot of emphasis in the user interface to be able to set that one attribute while not supporting the same level of configuration for other attributes (access, constant, etc.).  I think for the sake of simplicity it might be simpler to have a blanket "Ignore static" across all types for your use case as well as supporting turning off other attributes in the same fashion.  The existing option "Alphabetize members of the same group" is effectively the same style of option worded differently (i.e. it considers name, but name can be ignored by disabling).  A new option "Order static members of the same group above non-static members" enabled by default would fit your use case I think.  There's probably also some room to rework this options dialog, but that's some added scope and runs into backwards compatibility issues.

What do you think?

hyperquantum:
Any news about this? An option "ignore static" would be really useful.

shuelsmeier:
I still would prefer per type rules. As I already wrote, I would be willing to implement the other attributes (access, constant, etc.) as well.

Backwards compatibility issues related to existing configuration files are already covered by providing means of an upgrade mechanism (contained in the current pull request).

For example the reorganizing types dialog with methods opened/closed could look like this:

codemaid1

The types that are closed would show their configuration next to the type name.

codemaid2

hyperquantum:
Either way would be fine to me, but now this does not seem to make any progress towards getting merged, which is sad.

@shuelsmeier maybe you could implement the other attributes as you suggested and provide some way to make the settings easier for people who don't need per type rules, a "select/deselect all" of some sorts.

@codecadwallader would this be an acceptable compromise to you?

codecadwallader:
That's a good suggestion and I appreciate it.  My concern is that it seems like adding even more complexity to the configuration window for a use case that has not been requested.

I would still suggest a single check box that controlled static sorting across all types, similar to the existing check box for name sorting across all types.  I think this fits the original use case as well as minimizes the complexity of the configuration.

I definitely respect and understand a difference of opinions there.

Issue #698, opened 2019-12-30
Feature Request: Save Spade zoom level between sessions.


kmgallahan:
### Environment
- Visual Studio version: 2019
- CodeMaid version: 11.1.209

### Feature Request Description

It would be helpful if the Spade zoom level (set using the mousewheel) were saved between VS sessions.

Currently you have to re-zoom every time you close & open VS.

codecadwallader:
Thanks for the suggestion!

PR #699, opened 2019-12-30
Improving experience of Spade panel when dealing witgh large files

oleg-shilo:
Tis PR contains a couple of features to improve the user experience when dealing with overcrowded Spade panel of the very large files.

_This PR is a rework/reattempt of https://github.com/codecadwallader/codemaid/pull/671_

* Problem
Spade panel content does great job of showing the exact anatomy of the source code document. However that "exact" may not be always practical as often the spade content has very little code navigation value (e.g. back fields, indexers, ...). The problem becoms particulary difficult to deal with when deailng with the large files like autom-generated ViewModels, when the most of the realestate of the Spade paned is wasted on back fields and the really useful stuff like properties and methods cannot be accessed without scrolling, which needs to be repeated everytime user svitsch the active document to a new file.

* Solution

1. Similarly to the Properties panel of Visual Studio Spade panel now maintains collapsed tree view nodes even when active document is switched to a different file. 
The feature can be enabled/disabled from the user settings dialog. It is enabled by default.

2. User can completely hide some items' types in the tree view of Spade panel. 
The feature can be enabled/disabled from the user settings dialog. It is disabled by default.  

image

----

I also added suppressing UI-Thread warnings in the debug mode since they do not trigger any problems and you decided to not address them anyway (I actually agree with that approach). If it is not what you want it can be easily reverted in the project file.

I have also noticed that users requested some other Spade improvements associated with Spade panel remembering user choices (e.g. zoom) between the sessions: https://github.com/codecadwallader/codemaid/issues/698


codecadwallader:
Thanks again for putting this together. :)

In regards to keeping the same synchronized state across files, I see now that you are thinking of it like the properties tool window (i.e. it's the same set of categories independent of the files).  I think of it like each file is independent, because we synchronize the expand/collapse state to match the code editor.

image

With the code editor, if you collapse a region called "Fields" in File A, then switch to File B that same region will not be collapsed.  Each file has its own collapse/expand status.  We synchronize with the code editor by default so that if you switch between files, we can leverage the expand/collapse state that the code editor is remembering.



How do you think we can reconcile these two approaches?  It seems like maybe auto-collapsing some regions would be compatible and maybe meet your needs, but a second way to remember expand/collapse status could conflict with the existing way.  What do you think?

oleg-shilo:
>  It seems like maybe auto-collapsing some regions would be compatible and maybe meet your needs,

Important point, as I have already mentioned the feature I am proposing has nothing to do with code regions. Nothing at all. It is pure about nodes in the spade tree view. It is truly a small UX feature that does not interfere with with any existing functionality. Sorry if I did not make it clear before.

> I think of it like each file is independent, because we synchronize the expand/collapse state to match the code editor.

Sorry but I am not sure it is the case. When you have your spade categorized by member type the collapsed state of the nodes has no logical connection to the source code. The code has regions that are in no form or shape are lined to the nodes. There is no need (or possibility) to synchronize the collapsed  state of the spade node and code region. You can still synchronize the selection of the spade item and the caret in the code as you do now. I am not proposing to change anything in this area.

codecadwallader:
> Important point, as I have already mentioned the feature I am proposing has nothing to do with code regions. Nothing at all. It is pure about nodes in the spade tree view. It is truly a small UX feature that does not interfere with with any existing functionality. Sorry if I did not make it clear before.

Sorry for using the word "regions" as that caused confusion.  I did indeed mean the "nodes" in the Spade window.  The synchronization feature links those two to each other when sorting by line (default) but indeed they are two separate things.  Re-stated then:

It seems like maybe auto-collapsing some ~regions~ nodes would be compatible and maybe meet your needs,

> Sorry but I am not sure it is the case. When you have your spade categorized by member type the collapsed state of the nodes has no logical connection to the source code. The code has regions that are in no form or shape are lined to the nodes. There is no need (or possibility) to synchronize the collapsed state of the spade node and code region. You can still synchronize the selection of the spade item and the caret in the code as you do now. I am not proposing to change anything in this area.

I agree when sorting by alpha or type there is not a logical connection between them.  However when Spade is sorting by line (the default) it does have a logical connection to the source code and synchronizes expand/collapse states to match the regions.  This is done on a per-file basis.  Did the animated gif above help show that connection?

oleg-shilo:
Sorry Steve, I am a bit lost now.
> It seems like maybe auto-collapsing some regions nodes would be compatible
 
I assume you are talking about the feature `#1` ("remember expanded state of the nodes"). If it is the case and you indeed accept the feature then we are on the same page. Great.

> However when Spade is sorting by line (the default) it does have a logical connection to the source code.

Your statement is accurate. But let me try to see if I get it right. 
- When you implemented the non-default sorting you knew that it would break the code editor syncronization. Though despite that you let the user to decide to be (or not to be) comfortable with the price for the improved user experience. You gave the user a choice.

- When I implemented optional hiding some nodes I knew that it would break the code editor syncronization. Though despite that I let the user to decide to be (or not to be) comfortable with the price for the improved user experience. I gave the user a choice.

Sorry, I found it difficult to see the rationale for dealing with these two cases differently.

codecadwallader:
Thank you for trying to clarify the situation, I agree we seem to be thinking about it differently.

The original proposal was that expand/collapse state of nodes would behave like the properties window and not be tied to the file.  There is already existing functionality that is tied to the file.  Having the expand/collapse state of nodes be both file specific (current) and not file specific (proposed) within Spade seems confusing to me.  Does that make sense?

oleg-shilo:
>  Does that make sense?

Sorry, not much :)
I believe the fact that the new behavior is more consistent with IDE experiences is prevailing any potential confusions on the extension scope. 

I have to be honest, I found myself in a strange situation here. 

The very-very first contribution was based on remembering expand/collapse globally, unconditionally. It was rejected. The current proposal (file-based) is being rejected too. Doesn't leave me much room for negotiation.

I thought that if I show you that the existing behavior is not affected at all and encourage you to trust your users to decide for them what is confusing and what is not, then I would get you to support this... I am loosing my ground here too.

It's disappointing as it is my third unsuccessful attempt to contribute to this project. The project that I really like and even convinced the company to adopt it as a default formatting tool instead of stubborn ReSharper. 

Well, being an author of quite a few mainstream frameworks and IDE extensions I respect and fully accept your right to insist on your own vision of the product evolution.

Please feel free to reject and close this PR.

codecadwallader:
I'm sorry to hear that my explanations are not making sense.  It's unfortunate that we haven't come to a common understanding yet because I do genuinely appreciate the effort you've put forth and I want to be accepting.  I'm happy to keep discussing/iterating with you but I realize it has gone back and forth multiple times already and that can be frustrating.  Regardless, I want to say thank you for your efforts and for working to help improve CodeMaid.

Issue #659, opened 2019-06-21
[feature request] Add option to run new VS 2019 Sort usings command on save.


WarrenFerrell:
It would be nice to be able to the sort usings on save but not the "Sort and Remove" usings during automatic clean on save because the extremely annoying part is only really because of the removing usings,

codecadwallader:
Thanks for making the request.  I hadn't realized they re-introduced a "sort usings" only version of the command with VS2019.  There used to be one in older versions then it was taken away.

codecadwallader:
This is currently only in preview builds but will be cool when it gets mainstreamed.  We'll have to make it clear in the Options UI that it's only in VS2019.2+

https://visualstudiomagazine.com/articles/2019/06/11/vs-2019-16-2-update.aspx

PR #727, opened 2020-05-05
Fix Bug: RemoveAndSortUsings

kyleruddbiz:
Prevent duplicate using statements from being re-added.

This is somewhat related to #659, but I didn't see a bug describing this exact situation.

I'm including repro steps. Let me know if you would like me to move that out to a separate issue.

Visual Studio 2019 express. Windows 10.

1. Ensure CodeMaid Options match included image.
image
2. Cleanup Active Document.

Expected: Includes all referenced usings, and at most one instance of each using in CodeMaid options.
Actual: Reorders usings and adds duplicates.

[RemoveAndSortUsings Bug Example.zip](https://github.com/codecadwallader/codemaid/files/4583329/RemoveAndSortUsings.Bug.Example.zip)

### DIFF
--- a/CodeMaid/Logic/Cleaning/UsingStatementCleanupLogic.cs
+++ b/CodeMaid/Logic/Cleaning/UsingStatementCleanupLogic.cs
@@ -1,5 +1,4 @@
 using EnvDTE;
-using Microsoft.VisualStudio.Shell;
 using SteveCadwallader.CodeMaid.Helpers;
 using SteveCadwallader.CodeMaid.Properties;
 using System;
@@ -92,18 +91,20 @@ namespace SteveCadwallader.CodeMaid.Logic.Cleaning
                  .Where(usingStatement => TextDocumentHelper.FirstOrDefaultMatch(textDocument, string.Format(patternFormat, usingStatement)) == null)
                  .ToList();
 
-            ThreadHelper.ThrowIfNotOnUIThread();
-            var point = textDocument.StartPoint.CreateEditPoint();
-
-            foreach (string usingStatement in usingStatementsToReinsert)
+            if (usingStatementsToReinsert.Count > 0)
             {
-                point.StartOfLine();
-                point.Insert(usingStatement);
-                point.Insert(Environment.NewLine);
-            }
+                var point = textDocument.StartPoint.CreateEditPoint();
 
-            // Now sort without removing to ensure correct ordering.
-            _commandHelper.ExecuteCommand(textDocument, "Edit.SortUsings");
+                foreach (string usingStatement in usingStatementsToReinsert)
+                {
+                    point.StartOfLine();
+                    point.Insert(usingStatement);
+                    point.Insert(Environment.NewLine);
+                }
+
+                // Now sort without removing to ensure correct ordering.
+                _commandHelper.ExecuteCommand(textDocument, "Edit.SortUsings");
+            }
         }
 
         #endregion Methods
--- a/CodeMaid/Logic/Cleaning/UsingStatementCleanupLogic.cs
+++ b/CodeMaid/Logic/Cleaning/UsingStatementCleanupLogic.cs
@@ -1,4 +1,5 @@
 using EnvDTE;
+using Microsoft.VisualStudio.Shell;
 using SteveCadwallader.CodeMaid.Helpers;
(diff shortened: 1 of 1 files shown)


CodeMaid/Logic/Cleaning/UsingStatementCleanupLogic.cs
```diff
@@ -91,20 +85,25 @@ public void RemoveAndSortUsingStatements(TextDocument textDocument)
...
-                    point.editPoint.Insert(point.text);
-                    point.editPoint.Insert(Environment.NewLine);
-                }
+                point.StartOfLine();
+                point.Insert(usingStatement);
+                point.Insert(Environment.NewLine);
             }
+
+            // Now sort without removing to ensure correct ordering.
+            _commandHelper.ExecuteCommand(textDocument, "Edit.SortUsings");
```
codecadwallader, line 106: This will cause the usings to be sorted twice for newer versions of the IDE, even for the default case where the using statements to re-insert do not exist.  I can see why it is split out to happen after re-insertion, but I think it should be guarded to avoid extra work.

CodeMaid/Logic/Cleaning/UsingStatementCleanupLogic.cs
```diff
@@ -91,20 +85,25 @@ public void RemoveAndSortUsingStatements(TextDocument textDocument)
...
             }
 
-            // Check each using statement point and re-insert it if removed.
-            foreach (var point in points)
+            // Ignore any using statements that are still referenced
+            usingStatementsToReinsert = usingStatementsToReinsert
+                 .Where(usingStatement => TextDocumentHelper.FirstOrDefaultMatch(textDocument, string.Format(patternFormat, usingStatement)) == null)
+                 .ToList();
+
+            ThreadHelper.ThrowIfNotOnUIThread();
```
codecadwallader, line 95:
```suggestion
```
Although the IDE provides warnings about this, in general I've found we can invoke on background threads and boost performance.

CodeMaid/Logic/Cleaning/UsingStatementCleanupLogic.cs
```diff
@@ -73,16 +74,9 @@ public void RemoveAndSortUsingStatements(TextDocument textDocument)
...
-                          from editPoint in TextDocumentHelper.FindMatches(textDocument, string.Format(patternFormat, usingStatement))
-                          select new { editPoint, text = editPoint.GetLine() }).Reverse().ToList();
-
-            // Shift every captured point one character to the right so they will auto-advance
-            // during new insertions at the start of the line.
-            foreach (var point in points)
-            {
-                point.editPoint.CharRight();
-            }
+            var usingStatementsToReinsert = _usingStatementsToReinsertWhenRemoved.Value
```
codecadwallader, line 77: Previously I was trying to capture and re-insert text back where it was originally located.  This helped with some custom sort rules that users had.  I agree it may not be necessary anymore since newer versions of VS have merged remove and sort back together vs. being separate options.

CodeMaid/Logic/Cleaning/UsingStatementCleanupLogic.cs
kyleruddbiz: Good catch. I'll update code to skip that work if there are no reinsert candidates.

CodeMaid/Logic/Cleaning/UsingStatementCleanupLogic.cs
kyleruddbiz: Ah, that makes sense. I'll remove it.

CodeMaid/Logic/Cleaning/UsingStatementCleanupLogic.cs
kyleruddbiz: That makes sense, no problem.

kyleruddbiz:
Thanks for getting back to me. I made changes based on your comments. Let me know if you need any other adjustments.

codecadwallader:
Thanks again for putting this together and apologies for my delay in accepting it.  This looks great and ran well on my machine so I'm glad to merge it! :)

This bit of functionality has historically been a bit rough so I'm particularly excited to have it improved. <3

kyleruddbiz:
Happy to help, I look forward to the next update!

Issue #745, opened 2020-07-25
TextEditorEventListener.TextEditorEvents_LineChanged throws on every Output Window update


olegtk:
Visual Studio telemetry indicates one of the top UI delays is caused by this symbol:

stevecadwallader.codemaid.dll!SteveCadwallader.CodeMaid.Integration.Events.TextEditorEventListener.TextEditorEvents_LineChanged

The scenarios appears to be building large solution. On each Output Window update TextEditorEventListener.TextEditorEvents_LineChanged is called synchronously on the UI thread and the following line throws because there is no EnvDTE.Document corresponding to the Output Window (EnvDTE.Document represents document window):

var document = startPoint.Parent.Parent;

I suggest ignore Output Window and other non code files by checking first startPoint.Parent.Language - it will be "Plain Text" for the Output window and other txt files and text buffers hosted in other tool windows.

codecadwallader:
Thanks for reporting the issue and proposing a fix.  I haven't been able to reproduce yet that accessing `startPoint.Parent.Parent` causes an exception to be thrown.  I'm seeing it simply return `null` as expected and the following `if` guard exits out cleanly.  Can you please provide some more information (e.g. what version of VS, how you're getting the telemetry) so that I can look to reproduce it as well?

olegtk:
It's not a crash so I don't have a repro, but I thought it should be trivial to repro the exception by just running a build.
This issue is detected by Visual Studio PerfWatson as UI delay and it only captures UI thread. The delay affects all 16.0 versions seemingly, the rate is not that high, but for some reason significantly increased in 16.9.
Here is all I got:

clr.dll!CLREventWaitHelper2
clr.dll!CLREventWaitHelper
clr.dll!CLREventBase::WaitEx
clr.dll!SVR::gc_heap::user_thread_wait
clr.dll!WKS::gc_heap::background_gc_wait
clr.dll!WKS::gc_heap::try_allocate_more_space
clr.dll!WKS::gc_heap::allocate_large_object
clr.dll!WKS::GCHeap::Alloc
clr.dll!Alloc
clr.dll!SlowAllocateString
clr.dll!FramedAllocateString
mscorlib.dll!System.String.Concat
mscorlib.dll!System.Exception.InternalPreserveStackTrace
clr.dll!CallDescrWorkerInternal
clr.dll!CallDescrWorkerWithHandler
clr.dll!MethodDescCallSite::CallTargetWorker
[...]

PR #746, opened 2020-07-25
Ignore changes to Output Window

olegtk:
A proposed fix for #745


codecadwallader:
Thanks for putting this together.  It is pretty straight-forward but I'm wondering if there's a different check that should(tm) be made so I'd like to get it reproduced.  Let's continue the conversation in #745

PR #751, opened 2020-08-19
Allow CMake/Folder based projects to load extension.

obhi-d:
Hello,

Currently if you open CMake based projects, the extension does not load correctly, and you can only see the About and Options menu item. 
I added AutoLoad attribute for FolderOpened event, which fixes the issue. 
I let you decide if this can be merged.

Thanks,
obhi


obhi-d:
It seems the extension is not loaded even after adding this, unless one clicks on the Extension->CodeMaid menu item when the project is being loaded, I will try to find a fix.

ronan-fdev:
I spent hours trying to fix this (unsuccessfully). The main problem is that I have to uninstall Codemaid totally when working on a c-make project due to Intellisense stuck at "Wait for start". Disabling the extension is not enough....

PR #778, opened 2021-02-07
Feature: Cleanup - Only for some files

### LINKED ISSUE #773: Auto Cleanup On Save - Only for some files
### Environment
- Visual Studio version: 2019 Professional
- CodeMaid version: 11.1
- Code language: C#

### Description
I like the "Auto Cleanup On Save"-Feature to keep my files organized, but if I work in an old code-base I don't want to update every file where I might only changed one line, but would like to have all new added files auto cleaned on save.

I would propose something like a pragma that could be added to either all new files which should be auto cleanedup on save or to all files that should not auto cleaned on save. I would prefer the first solution.

Timo-Weike:
Introduces a new option to add a list of regular expressions that
determine which files should be included for clean up.

Fixes #773


codecadwallader:
This looks awesome!  Thanks so much for putting it together.

codecadwallader:
It's available in our CI build here https://www.vsixgallery.com/extension/4c82e17d-927e-42d2-8460-b473ac7df316 (v11.2.235 or higher)

Issue #790, opened 2021-03-01
File header update should actually update the existing header instead of simply adding one


lflender:
### Environment
- Visual Studio 2019
- CodeMaid version: 11.2
- Code language: Any

### Description
Currently, when a header is added in Cleaning/Update/File header, the provided sample is simply added to the file if the current header does not start with the sample. The current header should be entirely removed and replaced by the new one.

### Steps to recreate
1. Go to Cleaning/Update
2. Provide a file header, e.g. // Copyright © 2021 by MyCompany Inc.
3. Cleanup a file with an existing header, e.g. // Copyright © 2020 by MyCompany Inc.

### Current behavior
The new header is added before the previous one

### Expected behavior
The new header should replace the previous one

codecadwallader:
Thanks for reporting the issue. Please take a look at #757 and see if that helps.

lflender:
To me this is a different issue. I wish to replace old headers from files with a new one, and I cannot do that. New headers have a different structure from the old ones. CodeMaid currently does not manage the removal from any header and simply adds one.

Issue #757, opened 2020-10-08
File Header is inserted again, even if it already exists


VVander:
### Environment
- Visual Studio version: 2019 Pro
- CodeMaid version: 11.1
- Code language: C#

### Description
The file header is being inserted multiple times, even when it's not changed at all.
Probably related to [this issue](https://github.com/codecadwallader/codemaid/issues/310)

### Steps to recreate
1. "Enable Automatic Cleanup On Save" (General->Features) and "Run format comments during cleanup" (Formating)
2. Set the File Header for C# in the Cleaning->Update options to...
```
//%BANNER_BEGIN%
// ---------------------------------------------------------------------
//%COPYRIGHT_BEGIN%
//<copyright file="My File.cs" company="My Company">
//     Copyright (c) 2018-present, My Company, Inc. All Rights Reserved.
//</copyright>
//%COPYRIGHT_END%
// ---------------------------------------------------------------------
//%BANNER_END%
```
3. Save the file w/ ctrl+s and watch your header multiply!

### Current behavior
The header is being inserted again and again even though it's already there.

### Expected behavior
The header should only exist once.

codecadwallader:
Thanks for reporting the issue and for including your settings.  I was able to reproduce it and trace it down to trailing whitespace similar (but a little different) to #310

The second line of hyphens includes a trailing space.  Here's a screenshot of the settings with white space visible: 

![image](https://user-images.githubusercontent.com/3823772/95990099-16795400-0df9-11eb-8410-2ad2647e421d.png)

When cleanup runs, this end of line space is removed.. and then when cleanup next runs the copyright header is considered different and gets inserted again.  Deleting that space resolved the issue for me.

Hope it helps!

lolodi:
Hello! I noticed that if I save a file the header gets added correctly, but If run Cleanup All Code on a project the header is not added at all, while the rest of the cleanup is carried out as expected. Ideas?

codecadwallader:
Thanks for reporting the issue, I have been able to reproduce it.  I found this error message in diagnostics mode:

> InsertFileHeaderDocumentStart must be called on the UI thread.

It sounds like we need to add logic to marshal back onto the UI thread for this cleanup operation.

PR #797, opened 2021-03-12
File header update now replaces the file header if one is already present

lflender:
Fixing issues #790 #757

### DIFF
--- a/CodeMaid/Helpers/FileHeaderHelper.cs
+++ b/CodeMaid/Helpers/FileHeaderHelper.cs
@@ -1,0 +1,143 @@
+﻿using EnvDTE;
+using SteveCadwallader.CodeMaid.Properties;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text.RegularExpressions;
+
+namespace SteveCadwallader.CodeMaid.Helpers
+{
+    internal static class FileHeaderHelper
+    {
+        /// <summary>
+        /// Gets the file header from settings based on the language of the specified document.
+        /// </summary>
+        /// <param name="textDocument">The text document.</param>
+        /// <returns>A file header from settings.</returns>
+        internal static string GetFileHeaderFromSettings(TextDocument textDocument)
+        {
+            switch (textDocument.GetCodeLanguage())
+            {
+                case CodeLanguage.CPlusPlus: return Settings.Default.Cleaning_UpdateFileHeaderCPlusPlus;
+                case CodeLanguage.CSharp: return Settings.Default.Cleaning_UpdateFileHeaderCSharp;
+                case CodeLanguage.CSS: return Settings.Default.Cleaning_UpdateFileHeaderCSS;
+                case CodeLanguage.FSharp: return Settings.Default.Cleaning_UpdateFileHeaderFSharp;
+                case CodeLanguage.HTML: return Settings.Default.Cleaning_UpdateFileHeaderHTML;
+                case CodeLanguage.JavaScript: return Settings.Default.Cleaning_UpdateFileHeaderJavaScript;
+                case CodeLanguage.JSON: return Settings.Default.Cleaning_UpdateFileHeaderJSON;
+                case CodeLanguage.LESS: return Settings.Default.Cleaning_UpdateFileHeaderLESS;
+                case CodeLanguage.PHP: return Settings.Default.Cleaning_UpdateFileHeaderPHP;
+                case CodeLanguage.PowerShell: return Settings.Default.Cleaning_UpdateFileHeaderPowerShell;
+                case CodeLanguage.R: return Settings.Default.Cleaning_UpdateFileHeaderR;
+                case CodeLanguage.SCSS: return Settings.Default.Cleaning_UpdateFileHeaderSCSS;
+                case CodeLanguage.TypeScript: return Settings.Default.Cleaning_UpdateFileHeaderTypeScript;
+                case CodeLanguage.VisualBasic: return Settings.Default.Cleaning_UpdateFileHeaderVB;
+                case CodeLanguage.XAML: return Settings.Default.Cleaning_UpdateFileHeaderXAML;
+                case CodeLanguage.XML: return Settings.Default.Cleaning_UpdateFileHeaderXML;
+                default: return null;
+            }
+        }
+
+        internal static int GetHeaderLength(CodeLanguage language, string docStart)
+        {
+            switch (language)
+            {
+                case CodeLanguage.CSharp:
+                case CodeLanguage.CPlusPlus:
+                case CodeLanguage.JavaScript:
+                case CodeLanguage.LESS:
+                case CodeLanguage.SCSS:
+                case CodeLanguage.TypeScript:
+                    return GetHeaderLength(docStart, "//") +
+                        GetHeaderLength(docStart, "/*", "*/");
+
+                case CodeLanguage.HTML:
+                case CodeLanguage.XAML:
+                case CodeLanguage.XML:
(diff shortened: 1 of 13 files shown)


codecadwallader:
Thanks for putting this together.  Before diving into the code, one of the conceptual issues is how do you know the difference between an old header that should be replaced and a non-header comment that should be left intact?

Example:

```
// This is an old header that should be replaced
```

```
// This is a copyright statement from another package/library that should not be replaced
```

lflender:
Yeah I've thought about it and I don't think we can make it work for this particular case. I just take all leading comments and replace them with the new header, which is what we want 99% of the time. For edge cases, people would have to handle the issue manually.

Because when you take an existing repository which already has headers you wanna replace (which is the point of this PR), we fall in the exact same case you describe. So to me, it represents the vast majority of cases, which makes this feature useful. Trying to provide options and UI for particular cases is not worth it IMO and may be tedious for very little gain

lflender:
Or would you prefer an option to choose between update and insert?

codecadwallader:
That makes sense and I agree it may be the common case.  I do think though there's other cases where this could surprise people (e.g. differently styled comment header, commented out code) and we'd want to error on the side of not deleting potentially meaningful code.  Also it would be a breaking change with existing functionality which while it isn't very sophisticated it also isn't very harmful.

I think we should have the replace logic be off by default but be available as an opt-in option alongside the other file header options.  I think that has the advantage of not surprising existing users, but also enables users to improve the functionality with the changes you're proposing if they're comfortable with the risks.

lflender:
You are right, I added the option to choose between Insert or Replace, with Insert being the default.

lflender:
FYI my next PR will be to choose the placement of the header in C# files, before of after the usings block

CodeMaid/UI/Enumerations/HeaderUpdateMode.cs
```diff
@@ -0,0 +1,8 @@
+﻿namespace SteveCadwallader.CodeMaid.UI.Enumerations
+{
+    public enum HeaderUpdateMode
+    {
+        Insert,
```
codecadwallader:
```suggestion
        Insert = 0,
        Replace = 1
```
I find using explicit integer mappings for enums helps avoid mistakes where new values are added in the future and change values.

CodeMaid/Logic/Cleaning/FileHeaderLogic.cs
```diff
@@ -14,6 +15,8 @@ internal class FileHeaderLogic
 
         private readonly CodeMaidPackage _package;
 
+        private const int HeaderMaxNbLines = 30;
```
codecadwallader: Any rationale behind 30 lines or just needed an upper boundary?  Any thoughts on how to make this limit more transparent to users?

CodeMaid/Helpers/FileHeaderHelper.cs
```diff
@@ -0,0 +1,143 @@
...
+                if (!string.IsNullOrWhiteSpace(line))
+                {
+                    break;
+                }
+
+                header.Add(line);
+            }
+
+            // adds comment lines
+            foreach (var line in lines.ToList().Skip(header.Count))
```
codecadwallader, line 118: Would this logic be simpler if it simply counted any header lines and/or blank lines together?  I could see someone having a header format like:

```
// This is my header line

// There are many header lines but this one is mine
```

CodeMaid/Helpers/FileHeaderHelper.cs
```diff
@@ -0,0 +1,143 @@
...
+        {
+            if (!docStart.TrimStart().StartsWith(commentSyntaxStart))
+            {
+                return 0;
+            }
+
+            var startIndex = docStart.IndexOf(commentSyntaxStart);
+            var endIndex = docStart.IndexOf(commentSyntaxEnd);
+            var emptyStart = docStart.Substring(0, startIndex);
+            var emptyStartLength = emptyStart.Length - Regex.Matches(emptyStart, Environment.NewLine).Count;
```
codecadwallader: I'm not quite sure what this logic is handling.. it looks like its reducing the character length based on leading newlines?

CodeMaid/Logic/Cleaning/FileHeaderLogic.cs
```diff
@@ -69,32 +85,37 @@ internal void UpdateFileHeader(TextDocument textDocument)
...
+        {
+            var headerNbLines = Math.Min(HeaderMaxNbLines, textDocument.EndPoint.Line);
+            var headerBlockEnd = textDocument.StartPoint.CreateEditPoint();
+            
+            headerBlockEnd.MoveToLineAndOffset(headerNbLines, 1);
+
+            var headerBlockStart = textDocument.StartPoint.CreateEditPoint();
+            var headerBlock = headerBlockStart.GetText(headerBlockEnd);
+
+            return headerBlock;
```
codecadwallader: There is a `GetLines` method that will simplify this logic.  Something like:
```
var point = textDocument.StartPoint.CreateEditPoint();
var text = point.GetLines(1, headerNbLines);
```

codecadwallader (commented):
Thanks again for putting this together!  A few questions/comments when you have a chance.

CodeMaid/UI/Enumerations/HeaderUpdateMode.cs
lflender: fixed

CodeMaid/Logic/Cleaning/FileHeaderLogic.cs
lflender: No rationale, just looked fine to me. I just doubled it so it should be no problem, 60 lines is huge for a header.

I could add that as a tooltip somewhere but I really don't think users would benefit from this information. We could adjust it in case of bug report

CodeMaid/Helpers/FileHeaderHelper.cs
lflender: To me, headers are a single block. I add leading empty lines so this block doesn't have to begin at the start of the file. I was thinking about cases like that:

```
// This is my header block
//
// This is my header block
//
// This is my header block

// This is a comment about the below code that I want to keep
```

So I could change it if you like but this was intentional as I often see line breaks in comment blocks added with simple a "//" like in my example above.  You tell me

CodeMaid/Helpers/FileHeaderHelper.cs
lflender: You're right, my logic was flawed. I fixed and simplified it, fixed related unit tests and added comments and remarks trying to explain what's going on (counting only 1 character per end of line)

CodeMaid/Logic/Cleaning/FileHeaderLogic.cs
lflender: you're right, fixed

ghost:
- [ ]

lflender:
Is there an something wrong with my PR?

CodeMaid/Helpers/FileHeaderHelper.cs
codecadwallader: Ok, that makes sense.  I'm OK with assuming it has to be a contiguous section of lines without line breaks for now.

codecadwallader:
No your PR looks good.  I saw the checklist comment and thought it meant you were still working on it.  I see now that was posted by someone else, perhaps in accident.

I would like a little time next week to do some regression testing but the code looks good to me at this point.  Thanks for your efforts!

codecadwallader:
It worked great in my tests - nicely done!  Thanks again for making this contribution, it is very much appreciated. :)

PR #815, opened 2021-05-28
Added an option to update file headers after the "using" block in C#

lflender:
File headers can now be inserted or replaced by the one of the settings AFTER the block of "using" directive in C#

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,8 +6,10 @@ These changes have not been released to the Visual Studio marketplace, but (if c
 
 - [ ] Features
   - [x] [#778](https://github.com/codecadwallader/codemaid/pull/778) - Cleaning: Option to specify file inclusions by RegEx - thanks [Timo-Weike](https://github.com/Timo-Weike)!
+  - [x] [#797](https://github.com/codecadwallader/codemaid/pull/797) - Cleaning: Option to replace existing file headers (vs. default insert) - thanks [lflender](https://github.com/lflender)!
 
 - [ ] Fixes
+  - [x] [#800](https://github.com/codecadwallader/codemaid/pull/800) - Reorganizing: Fix dialog showing literal newline characters
 
 ## Previous Releases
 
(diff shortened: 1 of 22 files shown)


codecadwallader:
Thanks for putting this together.  I need some more time to review but everything looks good to me so far. 👍

codecadwallader:
@lflender I'm seeing an exception if I use the options replace and start of document together.

image
image

Will you please take a look?

lflender:
Sure, can you please give me the input file that raised this issue?

codecadwallader:
It was a default new class file like:

```
using System;
using System.Collections.Generic;

namespace ConsoleApp1
{
  class Program
  {
  }
}
```

lflender:
Hmmm, that's weird. I've tested all paths pretty thoroughly. I'll check what's wrong, give me a few days.

lflender:
I'm kinda confused here @codecadwallader, are you sure it has to do with what I changed? from what I see when I debug no part of the code failing has been modified in my changes.

If fails here
`TextDocumentHelper.InsertBlankLineBeforePoint(codeElement.StartPoint);`
in method
`internal void InsertPaddingBeforeCodeElements<T>(IEnumerable<T> codeElements)`
which is called by
`CodeCleanupManager.RunCodeCleanupCSharp`

It cannot access to codeElement.StartPoint for some reason. I don't see the relation with my changes, am I wrong?
Looks like a specific case where code cleanup fails, whatever the selected options are. It works fine otherwise.

codecadwallader:
Here's a zip file that includes the solution-specific settings that reproduce the issue: 

[ConsoleApp1.zip](https://github.com/codecadwallader/codemaid/files/6747898/ConsoleApp1.zip)

When I run against the `dev` branch it works as expected.  When I run against the PR branch it will throw the error as seen above.

I believe the issue is that if there is not a trailing newline in the settings then the comment is getting inserted and breaking the using statement.  I think the code previously would explicitly add a trailing newline if one was not present.

So for example if the setting is for:
```
// Comments
```
vs.
```
// Comments

```

Then when the new code performs the insertion it is coming out like:
```
// Comments using System.Collections.Generic;
```
which puts the code in a bad state (doesn't compile) and then breaks things.

Can you please take a look at the before/after code to see if you can find and make sure a trailing newline is still added?

lflender:
Thank you for the details. This is now fixed.

Sorry for the inconvenience!

codecadwallader:
Awesome, thank you!

lflender:
May I ask @codecadwallader  when you plan to publish the next release please? Approximately of course.

codecadwallader:
I would like to perform the next release "soon", but I would also like to spend some more time with the VS2022 preview and try to get it working so that can be released at the same time if possible.

Issue #762, opened 2020-11-18
Missing Microsoft.VSSDK.TestHostFramework nuget package


bgard68:
### Environment
- Visual Studio version: 2019 Enterprise
- CodeMaid version: latest from master
- Code language: C#

### Description
Missing Microsoft.VSSDK.TestHostFramework nuget package

### Steps to recreate
1. download repo or git clone repository
2. Open solution with visual studio 2019
3. Build the solution / install nuget packages
4.  Microsoft.VSSDK.TestHostFramework will not be found from nuget

### Current behavior
Cannot build the CodeMaid Integration Test Project

### Expected behavior
3 projects, including CodeMaid Integration Test Project should build successfully

codecadwallader:
Thanks for reporting the issue.  CodeMaid needs to be loaded in Visual Studio 2017 to work successfully.  I've learned over the years that for Visual Studio extensions to work correctly for all target versions it needs to be based on the _lowest_ version of Visual Studio that it supports (which is currently VS2017).

Some more information about getting up and started with compiling CodeMaid can be found here: http://www.codemaid.net/contribute/

Hope it helps!

PR #817, opened 2021-06-21
Enabling VS2022

eddynaka:


eddynaka:
There is an issue with this current PR: I can't find one dll to finish the migration:
Microsoft.VSSDK.TestHostFramework

I also didn't find any reference from it in the dev solution. Is that a nuget package? Is that a built-in from VS?

wertzui:
#762 says that it will work if you load it in VS2017

Cytraen:
`Microsoft.VSSDK.TestHostFramework.dll` is installed as part of VS, and in the 2022 Preview can be found in:
`C:\Program Files\Microsoft Visual Studio\2022\Preview\Common7\IDE\PublicAssemblies\`

Issue #816, opened 2021-06-17
Visual Studio 2022 Support


scottkuhl:
### Installed product versions
- Visual Studio: 2022 All Versions

### Description
Make the extension available in Visual Studio 2022.

### Steps to recreate
1. Download the extension installer.
2. Run the installer.

### Current behavior
Message that it is already installed to all applicable products appears even though it is not installed in Visual Studio 2022.
It is also not listed in Online Extensions in Manage Extensions Visual Studio 2022.

### Expected behavior
Installs and works in Visual Studio 2022.

_Note: Sorry for the all the extra details.  I probably could have stopped with the title on this one._

scottkuhl:
- [Writing Visual Studio Extensions with Mads - 64bit extensions for Visual Studio 2022](https://www.youtube.com/watch?v=L8b3kFriwCs)
- [Writing Visual Studio Extensions with Mads - Splitting extensions for Visual Studio 2022](https://www.youtube.com/watch?v=-PKIPTW6km0)

Eli-Black-Work:
Also looking for this 🙂

Eli-Black-Work:
Guess we can tell who the early adopters are, haha XD

wertzui:
There is already a PR for this #817

codecadwallader:
Thanks all for jumping on this quickly.  There are indeed some compatibility breaking changes in VS2022.  I believe last I heard it was expected that for all extensions the code would have to be branched and separate releases made targeting different versions of Visual Studio.  For the time being I would keep the `dev` and `release` branches pointing to the existing VS2017/VS2019 version, tied into the CI pipeline, etc.  We'll end up creating a new `vs2022` branch and I'm still thinking about the other implications.

It looks like 2 PRs have started, #817 and #818.  It looks like #818 might have gone farther with following recommendations.  Thanks for links @scottkuhl as there may be some non-obvious changes we need to make to support the changes.  Testing those out and seeing what issues come up (e.g. the settings dialog not opening in one) is super helpful and appreciated. <3

codecadwallader:
Hey all, I've been working on the VS2022 upgrade.  I decided to start fresh on a new `vs2022` branch and follow the instructions for multi-targeting VS versions based on https://docs.microsoft.com/en-us/visualstudio/extensibility/migration/update-visual-studio-extension?view=vs-2022 vs. building on top of #818 which is VS2022-only.  I have to say these are far more detailed instructions than have ever been provided before and I really appreciate them.

[...]

PR #818, opened 2021-06-23
VS 2022 support

Cytraen:
This PR migrates from `package.config` to `PackageReference` in the `.csproj` as recommended by Microsoft for VSIX projects targeting VS 2022+, in addition to targeting VS 2022. Both sets of tests build and the unit tests pass, but unfortunately I couldn't get the integration tests working due to `VSIDETestHostAdapter` being removed in VS 17.0 according to [this comment](https://developercommunity.visualstudio.com/t/integration-testshosttypevs-ide-are-not-worked-in/530262#T-N610979) on the Microsoft developer community board.

I couldn't get #817 to work for me, as opening the settings menu would throw an error, so this PR exists.


Cytraen:
I've been trying to split the extension for both VS 2017/2019 and VS 2022, as using the older SDK won't work for VS 2022, but as soon as I add a reference to a shared project all the XAML files effectively disappear. VS shows warnings that the `.xaml.cs` files are missing the `.xaml` files they depend on. Additionally, on trying to open the settings menu, I get `System.IO.IOException: 'Cannot locate resource 'ui/dialogs/options/optionswindow.xaml'.'` despite it being there in the `csproj`/`shproj` and all. Same issues happen if I move everything but the XAML and the corresponding code.

eduardocp:
@ashakoor Is the full path reference for `Microsoft.VSSDK.TestHostFramework` correct?

Cytraen:
> @ashakoor Is the full path reference for `Microsoft.VSSDK.TestHostFramework` correct?

Yes, I've just checked and it is correct.

Eli-Black-Work:
FYI, @codecadwallader has made a `vs2022` branch: https://github.com/codecadwallader/codemaid/issues/816#issuecomment-890166403🙂

CodeMaid.UnitTests/CodeMaid.UnitTests.csproj
```diff
@@ -108,11 +79,20 @@
       <Link>Properties\CodeMaid.snk</Link>
     </None>
     <None Include="app.config" />
-    <None Include="packages.config" />
   </ItemGroup>
   <ItemGroup>
-    <Analyzer Include="..\packages\Microsoft.VisualStudio.Threading.Analyzers.15.8.209\analyzers\cs\Microsoft.VisualStudio.Threading.Analyzers.CodeFixes.dll" />
-    <Analyzer Include="..\packages\Microsoft.VisualStudio.Threading.Analyzers.15.8.209\analyzers\cs\Microsoft.VisualStudio.Threading.Analyzers.dll" />
+    <PackageReference Include="Microsoft.VisualStudio.SDK">
+      <Version>17.0.0-previews-1-31410-273</Version>
```
RussKie, line 85: You could consider using [Microsoft.Build.CentralPackageVersions](https://github.com/microsoft/MSBuildSdks/blob/main/src/CentralPackageVersions/README.md) to avoid specifying same versions across multiple projects.
Here's an example of a similar migration: https://github.com/gitextensions/gitextensions/pull/9161

ekaraman89 (approved):

PR #828, opened 2021-07-28
Added an option to include access level in regions only for methods

lflender:


codecadwallader:
Thanks for putting this together, it looked good to me! 👍

PR #847, opened 2021-10-11
Replace usage of FindPattern/ReplacePattern with IFindService

olegtk:
Replace usage of FindPattern/ReplacePattern with IFindService.

DTE FindPattern and ReplacePattern are not supported in VS 2022, so this change replaces their usage with IFindService.

Addresses several issues listed in #816.


AmadeusW:
Looks good to me! CI shows build errors in a test project, perhaps they're missing the updated VSSDK dependency?

codecadwallader:
This is awesome - thanks for your hard work @olegtk !  I usually link all of the authors that go into each release in our blog posts, may I include you?

I see that IFindService is VS2019/VS2022 only.  We currently support VS2017 so I think we'll have to drop that with this release.  That's OK, we've had to do it from time to time (started with VS2005 support).

I'll play around with the `ThreadHelper.ThrowIfNotOnUIThread();` lines.  We deliberately try to do a lot of work on background threads for performance so this can be pretty painful to UX.  I've generally ignored those warnings although it makes me uneasy.

Thanks again! 👏

Issue #849, opened 2021-10-14
VS2022: Attempted to edit TextBuffer on the wrong thread


kfrancis:
### Environment
- Visual Studio version: VS2022
- CodeMaid version: CodeMaid.VS2022.20211014T145200 (see [this](https://github.com/codecadwallader/codemaid/issues/816#issuecomment-943462232))
- Code language: C#

### Description
Solution clean and getting these errors:

```
[CodeMaid Handled Exception 12:49:56 PM] CodeMaid Cleanup for 'ClaimAssertionBackchannelHandler.cs' was stopped: System.InvalidOperationException: Attempted to edit TextBuffer on the wrong thread.
   at Microsoft.VisualStudio.Text.Implementation.BaseBuffer.TextBufferBaseEdit..ctor(BaseBuffer baseBuffer)
   at Microsoft.VisualStudio.Text.Implementation.TextBuffer.CreateEdit(EditOptions options, Nullable`1 reiteratedVersionNumber, Object editTag)
   at Microsoft.VisualStudio.Text.Implementation.BaseBuffer.CreateEdit()
   at SteveCadwallader.CodeMaid.Helpers.TextDocumentHelper.ReplaceAll(ITextBuffer textBuffer, IEnumerable`1 replacements) in C:\src\codemaid\CodeMaidShared\Helpers\TextDocumentHelper.cs:line 385
   at SteveCadwallader.CodeMaid.Helpers.TextDocumentHelper.SubstituteAllStringMatches(TextDocument textDocument, String patternString, String replacementString) in C:\src\codemaid\CodeMaidShared\Helpers\TextDocumentHelper.cs:line 292
   at SteveCadwallader.CodeMaid.Logic.Cleaning.CodeCleanupManager.RunCodeCleanupCSharp(Document document) in C:\src\codemaid\CodeMaidShared\Logic\Cleaning\CodeCleanupManager.cs:line 272
   at SteveCadwallader.CodeMaid.Logic.Cleaning.CodeCleanupManager.<>c__DisplayClass19_0.<Cleanup>b__0() in C:\src\codemaid\CodeMaidShared\Logic\Cleaning\CodeCleanupManager.cs:line 170
   at SteveCadwallader.CodeMaid.Helpers.UndoTransactionHelper.Run(Action tryAction, Action`1 catchAction) in C:\src\codemaid\CodeMaidShared\Helpers\UndoTransactionHelper.cs:line 55
```

kfrancis:
[...]

PR #851, opened 2021-10-18
Fix threading exceptions when running Solution cleanup

olegtk:
Allow calling new methods (using IFindService) from background thread by switching to UI thread explicitly, which was done previously by DTE FindPattern/ReplacePattern methods implicitly (via COM marshalling).

Addresses #849.


codecadwallader:
Thanks for making these updates and the explanation!

Issue #879, opened 2021-11-22
`#endif` between method and its Attribute is getting deleted.


MineCake147E:
### Environment
- Visual Studio version: Community Version 17.1.0 Preview 1.1
- CodeMaid version: 12.0.300
- Code language: C#

### Description
I wrote this code:
```csharp
namespace CodeMaidBugRepro
{
    public class Class1
    {
#if DEBUG
        [Obsolete]
#endif
        public uint A(uint b) => b;
    }
}
```
After running the code cleanup with CodeMaid, the `#endif` gets removed, making build impossible.  

### Steps to recreate
1. Create a class library project.
2. Write a method.
3. Add `[Obsolete]` on top of the method. `[DebuggerStepThrough]` also gets affected.
4. Surround the attribute with `#if` and `#endif`.
5. Click on "Clean Up Active Documents" or turn on "Automatic Cleanup On Save" and save the document.

### Current behavior
The `#endif` gets removed.  
### Expected behavior
The `#endif` should remain alive.

codecadwallader:
Thanks for reporting the issue.  I have been able to reproduce it.  This looks to be fitting into a series of issues related to a breaking change in an underlying vsCMAccess API.  When we try to set access modifiers using the same API from previous versions it is now causing breaking changes to the code.

As a workaround, if you disable CodeMaid->Options->Cleaning->Insert->Insert explicit access modifier fields on methods that should(tm) avoid the issue for now.

codecadwallader:
I've confirmed this issue happens in both VS2019 and VS2022.  I've sent an email to Microsoft to try and get some help digging down into the API.

codecadwallader:
While looking in that area, there is also room for improvement in that `CodeElementHelper` doesn't recognize some syntax patterns (e.g. method arrow definitions) as well as `GetTextToFirstMatch` which was recently updated for VS2022 is not respecting the `startPoint`.  These issues are separate from the underlying API issue, but in the example above they would prevent the API from being called if the access modifier was already present.

WeihanLi:
Hi @codecadwallader , thanks for your working, any update on this?

codecadwallader:
Hi @WeihanLi , the latest update is I'm waiting for feedback from Microsoft on this issue if you would like to look/comment over there: https://github.com/dotnet/roslyn/issues/58315

flamewave:
I am also getting this issue, but it's removing the entire #if .... #endif block. For example:
```
[Attribute1]
[...]

Issue #866, opened 2021-11-10
Single-Line-Property not working on V12.0


Apflkuacha:
### Environment
- Visual Studio version: 2022 Community, also on 2019
- CodeMaid version: 12.0
- Code language: C#

### Description
With the new version, when the setting **Insert explicit access modifiers on properties** is activated, following line 

    public bool AdminMode { get { return _adminMode; } set { if (_adminMode != value) _adminMode = value; } }

is changed to 

    public bool AdminMode 
    { get { return _adminMode; } set { if (_adminMode != value) _adminMode = value; } }

This line should stay in one line. It stays in one line if the setting mentioned above is deactivated. Following line for example will stay in one line, independent of the setting:

    public string StatusMessages { get => _statusMessages; set => Set(ref _statusMessages, value); }

Off topic: why is there a separate 2022 Version of CodeMaid in the Visual Studio Marketplace? It was ranked that high in the default "Most Downloads" sort order, now it begins again with zero downloads?!

nbelley:
Same issue, it also happens in visual studio 2019. 

It took almost a week before we realized this... Some of our repos now need to be manually cleaned and this really sucks.

codecadwallader:
Thanks for reporting the issue.  It looks like a number of issues are cropping up around inserting explicit access modifiers on properties and the underlying API is no longer behaving the way it used to before.  At this point I would recommend turning off that option at CodeMaid->Options->Cleaning->Insert->Insert explicit access modifiers on->properties

@nbelley That does really suck, and I'm very sorry to hear it. :(

> Off topic: why is there a separate 2022 Version of CodeMaid in the Visual Studio Marketplace? It was ranked that high in the default "Most Downloads" sort order, now it begins again with zero downloads?!

VS2022 included some breaking changes and required all extension authors to split their extensions.  The recommendation was to create separate pages in the marketplace for each version.  Here's a couple links with more information:

https://www.codemaid.net/codemaid-v12-0-a-fork-in-the-road-edition-released/
https://docs.microsoft.com/en-us/visualstudio/extensibility/migration/update-visual-studio-extension?view=vs-2022

codecadwallader:
This lookup is not coming back with the expected results: https://github.com/codecadwallader/codemaid/blob/dev/CodeMaidShared/Logic/Cleaning/InsertExplicitAccessModifierLogic.cs#L297

[...]

PR #882, opened 2021-11-25
Fixed multiple bugs on file headers update

lflender:
- Fixed header replacement management: if a header is already present and a new one is added in another place using the "Replace" option, then the old one is removed.
e.g. if a header is present at document start and we add a new one after usings with the Replace option, the one at document start is now removed
- Header update on files with no header and no using is now working
- Header piece being sometimes inserted after header is now fixed

### DIFF
--- a/CodeMaid.UnitTests/Helpers/FileHeaderHelperTests.cs
+++ b/CodeMaid.UnitTests/Helpers/FileHeaderHelperTests.cs
@@ -1,4 +1,4 @@
-﻿using Microsoft.VisualStudio.TestTools.UnitTesting;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
 using NUnit.Framework;
 using SteveCadwallader.CodeMaid.Helpers;
 using System;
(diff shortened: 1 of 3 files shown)


codecadwallader:
Wonderful, thank you for making these fixes! :)

lflender:
Hey @codecadwallader, could I know when you plan to release the next version? I would like to use CodeMaid in my company with VS2022 and we need my latest changes. Maybe I could help with some bug fixes if you have set some objectives on that?

lflender:
I'm not used to milestones, just read about them... So a new version will be released after a number of PRs have been completed?

codecadwallader:
Hi @lflender , I don't have the next release planned yet.  Your changes should(tm) be available in the CI channel, but that's not as easy for everyone to install/update as an official release.  Some bug fix help would be great as there's definitely been an influx with the new VS2022 support!

The most urgent issue on my mind is that with VS2022 (and VS2019) the insert explicit access modifier APIs are misbehaving.  Several of the recently created issues (e.g. #879 #866) all tie back to this underlying issue.  I think if we had a solution for that it would definitely warrant a release as it would solve several issues.

I mainly use the milestones to link together issues into a release for posterity.  The idea being that if you wanted to see the changes in a specific release you could go to that milestone and see all the issues/PRs that were linked.

Thanks for your help!  Lets use #879 to communicate any progress on the insert explicit access modifier functionality.  That issue has the worst side effect IMO.

PR #905, opened 2022-02-03
Update GetTextToFirstMatch to use the start point as a basis for finding match

### LINKED ISSUE #867: CodeMaid reorders static new to new static
### Environment
- Visual Studio version: 2019 Professional
- CodeMaid version: 12.0
- Code language: C#

### Description
Codemaid reoders static new to new static and such conflicts with the microsoft recommendation, default, analyzer and .editorconfig.

### Steps to recreate
Reformat

    public static new App Current => (App)Application.Current;

and CodeMaid will change it to 

    public new static App Current => (App)Application.Current;

### Current behavior
The resort conflicts with the microsoft recommendation, default, analyzer and .editorconfig. So after codemaid is done the analyzer suggest a change in the document pointing to https://docs.microsoft.com/en-us/dotnet/fundamentals/code-analysis/style-rules/ide0036.

### Expected behavior
The order should follow Microsofts recommendation.

BlythMeister:
Update the logic to take into account the start point when searching.

This fixes #867

### DIFF
--- a/CodeMaid.config
+++ b/CodeMaid.config
@@ -11,7 +11,7 @@
         <value>\.Designer\.cs$||\.Designer\.vb$||\.resx$||CodeMaid.IntegrationTests\\.*\\Data\\||CodeMaid\\CodeMaid.cs||CodeMaid\\source.extension.cs</value>
       </setting>
       <setting name="Cleaning_AutoCleanupOnFileSave" serializeAs="String">
-        <value>True</value>
+        <value>False</value>
       </setting>
       <setting name="Reorganizing_RunAtStartOfCleanup" serializeAs="String">
         <value>False</value>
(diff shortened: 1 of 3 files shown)


codecadwallader:
Thank you very much for putting this together.  A couple thoughts:

1.  Can you please undo the common `CodeMaid.config` setting updates?
2.  When I pulled down this branch and utilized the code it did not seem to resolve #867 (the modifiers still flipped)?

BlythMeister:
Sorry, I didn't realise that I included that file.  That was an error.
I'll give it another look.  I tested on a method and it prevented the flip.  I didn't test on the exact code in that issue.  So will test with that too

BlythMeister:
@codecadwallader i've had a look and i think it's because it was an inline property and therefore had no `{` in it.
I've added `;` for struct and property to handle these being inline.

I've also reverted that accidentally commited file :)

i've now tested with:

```
namespace ConsoleApp1
{
    internal class Test
    {
        internal static new int MyValue => 1;

        internal static new void DoStuff()
        { }
    }
}
```

BlythMeister:
@codecadwallader is this PR ok now to be included in the next release?
Having the setting disabled has meant the valid auto adding of modifiers has also stopped which is a little frustrating.

codecadwallader:
Thanks for the updates!  I was able to confirm the fix now too.  Merging it in for the next release.  I really appreciate your help!

PR #968, opened 2023-01-14
Multi-bind on IsVisible to bypass a WPF leak issue

### LINKED ISSUE #966: BuildProgressView and RadialProgressBar leak WPF animations
Both BuildProgressView and RadialProgressBar leak WPF animations due to setting a WPF ProgressBar element's IsIndeterminate property to true. The value of this property needs to either be bound to the visibility of the control or be part of a multi-binding with its current logic + the visibility of the control.

This is causing leaked animations to show up in our internal report (I work on the Visual Studio team) as the number 8 and number 9 most common source of these leaks in 17.5

The WPF bug has been fixed here: https://github.com/dotnet/wpf/issues/6264

But that fix will not be ported back to the WPF version that VS uses, so individual offenders need to be fixed.

codecadwallader:
Proposed solution for #966

### DIFF
--- a/CodeMaidShared/CodeMaidShared.projitems
+++ b/CodeMaidShared/CodeMaidShared.projitems
@@ -159,6 +159,7 @@
       <AutoGen>True</AutoGen>
     </Compile>
     <Compile Include="$(MSBuildThisFileDirectory)UI\Bindable.cs" />
+    <Compile Include="$(MSBuildThisFileDirectory)UI\Converters\BooleanAndConverter.cs" />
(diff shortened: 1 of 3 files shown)


ryanmolden:
The fix looks good, will that also fix RadialProgressBar? It looks like it just hit BuildProgressView.

codecadwallader:
Thanks for taking a look, appreciated!

> I don't see that we are using the IsIndeterminate property on the RadilProgressBar, only the BuildProgressView. Does it have to be set even if it's not being used and is always false?

PR #1009, opened 2023-06-19
Fix compatibility problem with arm64 architecture

KevinJalais:
Hello,

I propose a few changes to make the extension compatible with VS 2022 installation on ARM64 architecture.
I have made the update of the NuGet packages and added the target "arm64" to the .vsixmanifest file.

The compilation is successful and I installed the extension to my VS 2022 running on Windows 11 virtualized with parallels desktop on a Macbook Pro (M2 Max chip) computer.

### DIFF
--- a/CodeMaid/CodeMaid.csproj
+++ b/CodeMaid/CodeMaid.csproj
@@ -340,12 +340,13 @@
       <Version>1.0.2</Version>
     </PackageReference>
     <PackageReference Include="Microsoft.VisualStudio.SDK">
-      <Version>16.10.31321.278</Version>
+      <Version>17.6.36389</Version>
       <ExcludeAssets>runtime</ExcludeAssets>
       <PrivateAssets>all</PrivateAssets>
+      <IncludeAssets>compile; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
     </PackageReference>
     <PackageReference Include="Microsoft.VSSDK.BuildTools">
-      <Version>16.11.35</Version>
+      <Version>17.6.2164</Version>
       <IncludeAssets>runtime; build; native; contentfiles; analyzers</IncludeAssets>
       <PrivateAssets>all</PrivateAssets>
     </PackageReference>
--- a/CodeMaid.UnitTests/CodeMaid.UnitTests.csproj
+++ b/CodeMaid.UnitTests/CodeMaid.UnitTests.csproj
@@ -81,25 +81,25 @@
   </ItemGroup>
   <ItemGroup>
     <PackageReference Include="Microsoft.VisualStudio.SDK">
-      <Version>16.10.31321.278</Version>
+      <Version>17.6.36389</Version>
     </PackageReference>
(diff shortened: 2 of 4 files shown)


CodeMaid/CodeMaid.csproj
```diff
@@ -340,12 +340,13 @@
       <Version>1.0.2</Version>
     </PackageReference>
     <PackageReference Include="Microsoft.VisualStudio.SDK">
-      <Version>16.10.31321.278</Version>
+      <Version>17.6.36389</Version>
```
codecadwallader, line 343: I believe these major package updates will make the extension incompatible with VS2019.

codecadwallader (commented):
Thank you for proposing some changes to add ARM support.  I believe some of these though will make the extension unable to compile for VS2019 since it needs to target the older SDK versions.  This has been a challenge with having a code base that spans multiple Visual Studio versions, we always have to target the oldest supported version.

KevinJalais:
Hi @codecadwallader, I understand your point.
Unfortunately, I am not able to test on VS2019 with Intel based computer.

goodcoffeecode:
This also works on the new Microsoft Surface devices that use the Snapdragon ARM64 architecture. I think this requirement might become more requested over the next 12 months as more manufacturers roll out Copilot+ devices.

burdittw:
You can load different packages using condition statements in your project files. For example in one of my projects I have the following to hide BuildHost-{framework} folders that get created if some libraries are multitargeted.
```
<ItemGroup Condition="'$(TargetFramework)' == 'net9.0'">
   <PackageReference Include="Microsoft.CodeAnalysis" Version="4.12.0" />
   <PackageReference Include="Microsoft.CodeAnalysis.Common" Version="4.12.0" />
</ItemGroup>
```
I am sure that you can use this same syntax to probably use the correct libraries based on the builds by a condition whatever that may be. Since I have no idea what that condition is I cannot provide any more details or even know if this will work. But this is how I take care of issues with multitargeted libraries in a project and this was provided to my by Microsoft support after upgrading to .NET 9.0 and I started getting the BuildHost folders in my web project.

Here is a link to the MS Learn site on using the Condition filter within the [ItemGroup](https://learn.microsoft.com/en-us/visualstudio/msbuild/itemgroup-element-msbuild?view=vs-2022) element. 

According to Copilot AI you can use the following to build based on Visual Studio version:

> To create an ItemGroup Condition in Visual Studio 2022 that differentiates builds based on whether Visual Studio 2019 or Visual Studio 2022 is being used, you can leverage the `$(VisualStudioVersion)` property within the condition, checking if it's greater than or equal to "17.0" (which represents Visual Studio 2022): `<ItemGroup Condition="'$(VisualStudioVersion)' >= '17.0'">`; this will only include the listed items in the ItemGroup when building with Visual Studio 2022 or later. 

This should allow you to add only packages by specific versions per Visual Studio Versions so you can I would assume is multiple versions of the files based on VS Version so everyone can get what they need and you only have one set of code.

